PR #5, opened 2020-07-15
Switch from GraphQL to gRPC endpoint

edigaryev:
This needs an RPC endpoint serving over TLS to make things secure.


pkg/parser/parser.go
```diff
@@ -1,121 +1,77 @@
...
-	if err != nil {
-		return nil, fmt.Errorf("%w: %v", ErrTransport, err)
+	request := api.ParseConfigRequest{
+		Config: config,
 	}
 
-	// Parse and validate JSON result
-	var parsed gqlResponse
-	err = json.Unmarshal(responseBody, &parsed)
+	_, err = client.ParseConfig(ctx, &request)
```
fkorotkov, line 60: The response now contains parsed tasks. Maybe we can follow up this PR with one to have a flag to print these tasks out? Or maybe globally support `--verbose` flag to have verbose logging of everything. 🤔

edigaryev: That would be a nice and logical addition indeed, however it seems that these details would be more applicable to the command that actually executes the tasks (e.g. `cirrus run`). Perhaps we should implement it first, and then the answer will be more evident?

fkorotkov:
https://grpc.cirrus-ci.com should be more or less stable now. Let's see how it goes.

PR #6, opened 2020-07-21
Basic implementation of the executor and the run command

edigaryev:
With these changes you can _try to_ self-build the CLI `go run cmd/cirrus/main.go run -v` on macOS.

Some things that still clearly need to be done:

* enable Docker-in-Docker in `.cirrus.yml`
* make `host.docker.internal` work on Linux
* create more elaborate tests
* handle termination signals and do a proper cleanup


internal/executor/rpc/rpc.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if err := stream.SendAndClose(&api.UploadLogsResponse{}); err != nil {
+		r.logger.WithContext(stream.Context()).WithError(err).Warn("while closing log stream")
+		return err
+	}
+
+	return nil
+}
+
+func (r *RPC) Heartbeat(ctx context.Context, req *api.HeartbeatRequest) (*api.HeartbeatResponse, error) {
```
fkorotkov, line 221: Let's log a heartbeat here. Can be useful.

internal/executor/build/task.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/cirruslabs/cirrus-cli/internal/executor/instance"
+	"strconv"
+	"time"
+)
+
+type TaskStatus int
+
+const (
+	StatusNew TaskStatus = iota
+	StatusSucceeded
```
fkorotkov, line 15: Let's move it in a separate package since `task.Status = build. StatusSucceeded` looks a bit weird.

internal/executor/build/build.go
```diff
@@ -0,0 +1,101 @@
...
+	ProjectDir string
+
+	// Environment variables that will be injected via an agent
+	Environment map[string]string
+
+	// The actual tasks comprising this build
+	tasks map[int64]*Task
+
+	// A mutex to guarantee safe access to this build from both the "main loop" and gRPC server handlers
+	Mutex sync.Mutex
```
fkorotkov, line 23: I haven't seen any build access pattern that was not thread safe. I think it will be safe to remove Mutex. Did I miss something? Once we have parallel tasks and a fancy cmd UI then we'll need some synchronization there.

edigaryev: I don't like this either, but this is not just about the locking per se, more about the memory barriers and visibility.

Without following the intricate rules of the [Go memory model](https://golang.org/ref/mem) (which is not recommended, see "Advice" section) or implementing an explicit synchronization with e.g. `sync.Mutex`, there's no guarantee that we'll see up-to-date task status values in `Executor.Run()`, since gRPC server that we also run handles requests and updates status values [in separate goroutines](https://stackoverflow.com/questions/50365652/does-the-grpc-server-use-one-new-goroutine-to-process-one-method-call), which easily translates to "runs in a separate thread", which in turn [can easily cause problems](https://en.wikipedia.org/wiki/Memory_barrier#Example).

internal/executor/build/task.go
edigaryev: Indeed! Take a look at https://github.com/cirruslabs/cirrus-cli/pull/6/commits/9536927d0e93c7782b6f885dea6625382831a18f.

The other option is to move the whole `Task` to a separate `task` package, but that would clash a lot more with the variable named `task` across the codebase and would probably require us to rename the import from `task` to `taskpkg` or `btask`.

internal/executor/rpc/rpc.go
edigaryev: Added in https://github.com/cirruslabs/cirrus-cli/pull/6/commits/7fafcd8aaf9c132b30316712571baabc6ee9edc6.

internal/executor/instance/instance.go
```diff
@@ -0,0 +1,143 @@
...
+	logger.WithContext(ctx).Debugf("starting container %s", cont.ID)
+	if err := cli.ContainerStart(ctx, cont.ID, types.ContainerStartOptions{}); err != nil {
+		return err
+	}
+
+	logger.WithContext(ctx).Debugf("waiting for container %s to finish", cont.ID)
+	waitChan, errChan := cli.ContainerWait(ctx, cont.ID, container.WaitConditionNotRunning)
+	select {
+	case res := <-waitChan:
+		logger.WithContext(ctx).Debugf("container exited with %v error and exit code %d", res.Error, res.StatusCode)
```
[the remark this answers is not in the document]
edigaryev: This is a debug level since it happens pretty much every time (even if the container exited cleanly). So, even if the actual task failed — this will still report exit code 0, because that is an exit code for an agent, not the actual task.

And since we're running tasks through an intermediate layer in the form of the agent — the actual "exit" is conveyed in the `ReportSingleCommand` RPC call (or via instance timeout in the main loop).

You think this would still make sense if we use `Warnf` here?

internal/executor/instance/instance.go
```diff
@@ -0,0 +1,143 @@
...
+	if instance.Type != "container" {
+		return nil, ErrUnsupportedInstance
+	}
+	var taskContainer api.ContainerInstance
+	if err := proto.Unmarshal(instance.Payload, &taskContainer); err != nil {
+		return nil, err
+	}
+
+	return &Instance{
+		image: taskContainer.Image,
```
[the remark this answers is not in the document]
edigaryev: I've created an issue for this (https://github.com/cirruslabs/cirrus-cli/issues/7) to make it more actionable.

internal/executor/instance/instance.go
```diff
@@ -0,0 +1,143 @@
...
+
+type Instance struct {
+	image string
+}
+
+var ErrUnsupportedInstance = errors.New("unsupported instance type")
+
+const (
+	// ContainerProjectDir specifies where in the instance container should we bind-mount the projectDir.
+	ContainerProjectDir = "/tmp/cirrus-ci/project-dir"
```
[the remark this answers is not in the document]
edigaryev: Perhaps you meant `/tmp/cirrus-ci/working-dir` here?

`ContainerProjectDir` is where the host's directory with `.cirrus.yml` inside is mounted in read-only mode.

internal/executor/agent/agent.go
```diff
@@ -0,0 +1,201 @@
...
+
+	err = cli.ContainerRemove(ctx, containerName, types.ContainerRemoveOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createAgentVolume(
```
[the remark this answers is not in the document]
edigaryev: Do you have any convention or a coding style document in mind that specifically addresses this? 🤔

It seems that by changing this we're essentially introducing our own coding style which will complicate drive-by contributions (simply because it needs to be communicated via a written document, lint rule or a PR comment).

On top of that, this adds unnecessary cognitive load to the developer when making a change (due to ambiguities in determining which function should follow which), yet adds a little benefit with the advent of modern IDE's that make sure there's no practical difference between any of the approaches.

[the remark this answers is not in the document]
fkorotkov: That's just how I always wrote code. Don't remember who taught me that but just check how `GetAgentVolume` ends:

```go
return createAgentVolume(ctx, cli, containerNameTemporary, containerNameFinal, volumeName)
```

It will be very nice for human readability to see `createAgentVolume` right after it. We can investigate if golangci-lint has rule for it and if not we can try to implement it.

internal/executor/build/build.go
fkorotkov: Yes, task's status is the only thing that changes and CLI need to have updated value for figuring out which task to run next. We can just use some atomic in order to do the updates (make status private and use getter and setter for the atomic operation). What do you think?

internal/executor/instance/instance.go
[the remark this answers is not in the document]
fkorotkov: `$TEMP/cirrus-ci-build` is the default folder the agent tries to run a task in. See https://github.com/cirruslabs/cirrus-ci-agent/blob/fff5c87cefa66c4a10663cc1fb59b32cc1e46c01/internal/executor/executor.go#L163

internal/executor/instance/instance.go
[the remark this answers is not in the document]
fkorotkov: Let's keep debug level if it happens pretty much every time.

internal/testutil/fs.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+// tempDir supplements an alternative to TB.TempDir()[1], which is only available in 1.15.
+// [1]: https://github.com/golang/go/issues/35998
+func TempDir(t *testing.T) string {
+	dir, err := ioutil.TempDir("", t.Name())
```
fkorotkov, line 12: Let's make it run on Mac with the default configuration:

```suggestion
	tempDirRoot := "" // will use os.TempDir()
	if runtime.GOOS == "darwin" {
		// override the default since Docker for Mac
		// doesn't mount /var/folder by default where os.TempDir() will be located
		// See https://docs.docker.com/docker-for-mac/#file-sharing
		tempDirRoot = "/tmp"
	}

	dir, err := ioutil.TempDir(tempDirRoot, t.Name())
```

internal/testutil/fs.go
```diff
@@ -0,0 +1,31 @@
+package testutil
+
+import (
+	"io/ioutil"
+	"os"
```
fkorotkov, line 5: Will need an import

```suggestion
	"os"
	"runtime"
```

internal/executor/build/build.go
edigaryev: I thought about wrapping it in an interface method too, but don't you think that in the near future more fields will show up, and it would basically end up with the same mutex wraps?

fkorotkov: I just don't like concurrency primitives except channels and atomics. In most of the cases use of a Mutex overcomplicates things. I'll think how we can re-architecture things like `instance`, `executor` and `build`. 🤔

fkorotkov: On the second though it seems we'll have serialization of the status update by nature of how things are executed at the moment: a status is updated **before container is killed** and the access to the task status happens **after container is cleaned up** in order to start a new task.

I'm currently also don't like how `instance` is responsible for executing itself. It simplifies things when we are running one task at a time but with the parallel execution we'll need to figure out some sort of instance manager that will be aware of local resources available to the CLI and run as many tasks in parallel as possible and there will need synchronization. Probably with some sort of a channel with task updates.

edigaryev: > On the second though it seems we'll have serialization of the status update by nature of how things are executed at the moment: a status is updated **before container is killed** and the access to the task status happens **after container is cleaned up** in order to start a new task.

You're right, but The Go Memory Model document clearly [recommends against this](https://golang.org/ref/mem#tmp_1):

>Programs that modify data being simultaneously accessed by multiple goroutines must serialize such access.
>
>To serialize access, protect the data with channel operations or other synchronization primitives such as those in the sync and sync/atomic packages.
>
>If you must read the rest of this document to understand the behavior of your program, you are being too clever.
>
>Don't be clever.

I'll implement the getter/setter methods as discussed above, and if you will keep the history while merging this, we will be able to reference the `sync.Mutex` solution any time in the future (props to Git).

>I'm currently also don't like how `instance` is responsible for executing itself.

Is it? As far as I can see, it is being executed by the `Executor` by calling the `Run()` method on an `Instance`.

>It simplifies things when we are running one task at a time but with the parallel execution we'll need to figure out some sort of instance manager that will be aware of local resources available to the CLI and run as many tasks in parallel as possible and there will need synchronization. Probably with some sort of a channel with task updates.

I agree that an instance manager makes sense in the future, but hey, this is a basic implementation just so we can start moving forward!

fkorotkov: Ok, so we are agreeing on moving the mutex to task and having getter / setter? I'll be OK with that and we can revisit it once we'll think about the parallel execution.

edigaryev: Added `sync.Mutex`-wrapped getter and setter for the task's state field in https://github.com/cirruslabs/cirrus-cli/pull/6/commits/a9a2096067d4dab3ef6fb866d95ea0cc465bb3ba.

internal/executor/build/task.go
```diff
@@ -0,0 +1,90 @@
...
+	ID       int64
+	status   taskstatus.Status
+	Instance *instance.Instance
+	Timeout  time.Duration
+
+	// Original Protocol Buffers structure for reference
+	ProtoTask *api.Task
+
+	// A mutex to guarantee safe accesses from both the main loop and gRPC server handlers
+	Mutex sync.Mutex
```
fkorotkov, line 25: If we are going the mutex way let's use `RWMutex` here.

internal/executor/agent/agent.go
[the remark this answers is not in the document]
fkorotkov: Another point here is to declare public functions first in the file.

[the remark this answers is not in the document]
edigaryev: See 8b1d285e87f0ba7266803329bac08ace47ce8b49.

internal/executor/build/task.go
edigaryev: See dce79d79cd0a1469b4605d36e2fabb8d64703f71.

edigaryev:
>enable Docker-in-Docker in `.cirrus.yml`

https://github.com/cirruslabs/cirrus-cli/pull/6/commits/0489a142a599184d1ec92f11fb277cb83f7eb3b5

> * make `host.docker.internal` work on Linux

https://github.com/cirruslabs/cirrus-cli/pull/6/commits/05ad37b8d513bd3e29986fdd3a9acb782f178b0a

> * create more elaborate tests

https://github.com/cirruslabs/cirrus-cli/pull/6/commits/56ed0150c23d089168171d68f70a7c5dbe0c8d53, https://github.com/cirruslabs/cirrus-cli/pull/6/commits/804c5d7231923adf73705a9d0a8d117b7009c6ca, https://github.com/cirruslabs/cirrus-cli/pull/6/commits/293e42995d4f8af0cc6dcc3192100167a1f3eee5, https://github.com/cirruslabs/cirrus-cli/pull/6/commits/d0ab9a973ca1c633aca4ea99e349779e537e2597, https://github.com/cirruslabs/cirrus-cli/pull/6/commits/9fe3bd26b986702187682aa84017b5aa49a24a74

> * handle termination signals and do a proper cleanup

https://github.com/cirruslabs/cirrus-cli/pull/6/commits/47aff47c40439e8a1402d327b24a5bc2256158bd

internal/executor/rpc/rpc.go
```diff
@@ -0,0 +1,310 @@
...
+	logEntry := r.logger.WithFields(map[string]interface{}{
+		"task":    task.ProtoTask.Name,
+		"command": req.CommandName,
+	})
+
+	// Register whether the current command succeeded or failed
+	// so that the main loop can make the decision whether
+	// to proceed with the execution or not.
+	if req.Succeded {
+		logEntry.Debug("command succeeded")
```
fkorotkov, line 202: Let's add `CommandName` to the log message

internal/executor/rpc/rpc.go
```diff
@@ -0,0 +1,310 @@
...
+
+		nextCommand = getCommandAfter(task.ProtoTask.Commands, req.CommandName)
+		if nextCommand == "" {
+			task.SetStatus(taskstatus.Succeeded)
+		}
+	} else {
+		logEntry.Debug("command failed")
```
fkorotkov, line 209: Let's add `CommandName` to the log message

internal/executor/rpc/rpc.go
edigaryev: It's already in the log message, see the full change in the commit that introduces this (https://github.com/cirruslabs/cirrus-cli/pull/6/commits/1f89bcb54a9ab99f7f97b2c3d23fa7aec1ddb27d).

PR #12, opened 2020-07-28
Output logs by default

fkorotkov:
Since we are not running tasks in parallel yet we can simply output logs for troubleshooting.


internal/executor/build/task.go
```diff
@@ -43,7 +43,10 @@ func NewFromProto(protoTask *api.Task) (*Task, error) {
 			Name: command.Name,
 			Instruction: &api.Command_ScriptInstruction{
 				ScriptInstruction: &api.ScriptInstruction{
-					Scripts: []string{"cp -rT $CIRRUS_PROJECT_DIR ."},
+					Scripts: []string{
+						"echo 'Coping mounted project to a working directory...'",
```
edigaryev, line 47:
```suggestion
						"echo 'Copying mounted project directory to a working directory...'",
```

PR #15, opened 2020-07-29
Support additional containers

edigaryev:
[Port mappings](https://cirrus-ci.org/guide/writing-tasks/#additional-containers) are not supported for now and the warning message is printed for each additional container that uses a port mapping.

### DIFF
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -150,24 +150,24 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
 	}
 
 	// Start additional containers (if any)
-	var acWG sync.WaitGroup
-	acErrChan := make(chan error, len(inst.additionalContainers))
-	acCtx, acCancel := context.WithCancel(context.Background())
+	var additionalContainersWG sync.WaitGroup
+	additionalContainersErrChan := make(chan error, len(inst.additionalContainers))
+	additionalContainersCtx, additionalContainersCancel := context.WithCancel(context.Background())
 	for _, additionalContainer := range inst.additionalContainers {
 		additionalContainer := additionalContainer
 
-		acWG.Add(1)
+		additionalContainersWG.Add(1)
 		go func() {
-			if err := runAdditionalContainer(acCtx, logger, additionalContainer, cli, cont.ID); err != nil {
-				acErrChan <- err
+			if err := runAdditionalContainer(additionalContainersCtx, logger, additionalContainer, cli, cont.ID); err != nil {
+				additionalContainersErrChan <- err
 			}
-			acWG.Done()
+			additionalContainersWG.Done()
 		}()
 	}
 
 	defer func() {
-		acCancel()
-		acWG.Wait()
+		additionalContainersCancel()
+		additionalContainersWG.Wait()
 
 		logger.WithContext(ctx).Debugf("cleaning up container %s", cont.ID)
 		err := cli.ContainerRemove(context.Background(), cont.ID, types.ContainerRemoveOptions{Force: true})
@@ -188,7 +188,7 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
 		logger.WithContext(ctx).Debugf("container exited with %v error and exit code %d", res.Error, res.StatusCode)
 	case err := <-errChan:
 		return err
-	case acErr := <-acErrChan:
+	case acErr := <-additionalContainersErrChan:
 		return acErr
 	}
 
@@ -198,12 +198,12 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
 func runAdditionalContainer(
 	ctx context.Context,
 	logger *logrus.Logger,
-	ac *api.AdditionalContainer,
+	additionalContainer *api.AdditionalContainer,
 	cli *client.Client,
 	connectToContainer string,
 ) error {
-	logger.WithContext(ctx).Debugf("pulling image %s", ac.Image)
-	progress, err := cli.ImagePull(ctx, ac.Image, types.ImagePullOptions{})
+	logger.WithContext(ctx).Debugf("pulling image %s", additionalContainer.Image)
+	progress, err := cli.ImagePull(ctx, additionalContainer.Image, types.ImagePullOptions{})
 	if err != nil {
 		return fmt.Errorf("%w: %v", ErrAdditionalContainerFailed, err)
 	}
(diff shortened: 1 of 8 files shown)


internal/executor/instance/instance.go
```diff
@@ -126,12 +133,42 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
...
+
 	cont, err := cli.ContainerCreate(ctx, &containerConfig, &hostConfig, nil, "")
 	if err != nil {
 		return err
 	}
 
+	// Start additional containers (if any)
+	var acWG sync.WaitGroup
+	acErrChan := make(chan error, len(inst.additionalContainers))
+	acCtx, acCancel := context.WithCancel(context.Background())
```
fkorotkov, line 155: Should we use `ctx` here so parent cancelation is handeled too?

```suggestion
	acCtx, acCancel := context.WithCancel(ctx)
```

internal/executor/instance/instance.go
```diff
@@ -126,12 +133,42 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
...
+		containerConfig.Env = append(containerConfig.Env, "CIRRUS_PORTS_WAIT_FOR="+commaDelimitedPorts)
+	}
+
 	cont, err := cli.ContainerCreate(ctx, &containerConfig, &hostConfig, nil, "")
 	if err != nil {
 		return err
 	}
 
+	// Start additional containers (if any)
+	var acWG sync.WaitGroup
```
fkorotkov, line 153: Maybe use a more descriptive name like `additionalContainersWaitGroup` or `additionalContainersWG`?

internal/executor/instance/instance.go
```diff
@@ -151,7 +188,86 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
...
+		return acErr
 	}
 
 	return nil
 }
+
+func runAdditionalContainer(
+	ctx context.Context,
+	logger *logrus.Logger,
+	ac *api.AdditionalContainer,
```
fkorotkov, line 201: maybe also use a more descriptive name like `additionalContainer`?

internal/executor/instance/instance.go
edigaryev: Looks to verbose to me. Do you want to rename `acCtx` and `acCancel` too?

internal/executor/instance/instance.go
edigaryev: It's handled through `acCancel()` in `defer`, which we'll need to call anyway in case of successful termination.

Another reason is that using a separate context here makes cancellation order more predictable, which helps when reasoning about `Instance.Run()`'s properties.

internal/executor/instance/instance.go
fkorotkov: Yeah, this is by JVM background talks 😅 But Go overall is very verbose! In general if I see a one or two letters variable beside `i` and `j` I want to understand what it is without looking up the declaration. `WG` seems like a convention thought but personally I'd name it `additionalContainersWaitGroup`.

internal/executor/instance/instance.go
fkorotkov: I don't see how `Interrupt` will be handled in this case. You've added `ctx` for that reason, don't you?

internal/executor/instance/instance.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/15/commits/f0d70b442879f54bfcae1e3c1b94fde9a1c34b4e.

internal/executor/instance/instance.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/15/commits/f0d70b442879f54bfcae1e3c1b94fde9a1c34b4e.

internal/executor/instance/instance.go
edigaryev: > I don't see how `Interrupt` will be handled in this case.

https://github.com/cirruslabs/cirrus-cli/blob/f0d70b442879f54bfcae1e3c1b94fde9a1c34b4e/internal/executor/instance/instance.go#L168-L169

internal/executor/instance/instance.go
```diff
@@ -150,24 +150,24 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
 	}
 
 	// Start additional containers (if any)
-	var acWG sync.WaitGroup
-	acErrChan := make(chan error, len(inst.additionalContainers))
-	acCtx, acCancel := context.WithCancel(context.Background())
+	var additionalContainersWG sync.WaitGroup
```
fkorotkov, line 153: Thank you! 🙌

internal/executor/instance/instance.go
fkorotkov: I don't believe there is any guarantees that a `defer`ed function will be called on cancelation.

fkorotkov: So the `ctx` will get canceled, then `errChan` from `ContainerWait` will emit an error which then return from the function and finally the `defer`ed function will be called.

fkorotkov: Let' add a comment then: `// separate context so when 'ctx' is canceled the main container will be killed first and only then all additional containers will be killed too`

internal/executor/instance/instance.go
```diff
@@ -126,12 +133,42 @@ func (inst *Instance) Run(ctx context.Context, config *RunConfig) error {
...
+		go func() {
+			if err := runAdditionalContainer(additionalContainersCtx, logger, additionalContainer, cli, cont.ID); err != nil {
+				additionalContainersErrChan <- err
+			}
+			additionalContainersWG.Done()
+		}()
+	}
+
 	defer func() {
+		additionalContainersCancel()
```
fkorotkov, line 169: With that logic I think cancalation of the additional containers context should be after removing the main container. 

What do you think about moving this defer right after declaration of `additionalContainersCancel`?

edigaryev: > With that logic I think cancalation of the additional containers context should be after removing the main container.

See https://github.com/cirruslabs/cirrus-cli/pull/15/commits/467c8f3159f558764b43a5c90c891c5700912607.

> What do you think about moving this defer right after declaration of `additionalContainersCancel`?

See https://github.com/cirruslabs/cirrus-cli/pull/15/commits/b67ac6007febf1b02e2a5e1fc47758638cb03d0a.

internal/executor/instance/instance.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/15/commits/1e7c3b93b9211257d4d5f91943b06d6b45ca0c22.

PR #17, opened 2020-08-03
Cache support

edigaryev:
This implementation attempts to support parallel CLI executions and has no automatic cleaning for now.

### DIFF
--- a/internal/executor/cache/cache.go
+++ b/internal/executor/cache/cache.go
@@ -32,7 +32,7 @@ func New(dir string, namespace string) (*Cache, error) {
 
 	// Create a base directory, ignoring ErrExist since it may already be created
 	// by a previous or parallel invocation of the CLI
-	if err := os.Mkdir(namespaceDir, 0700); err != nil {
+	if err := os.MkdirAll(namespaceDir, 0700); err != nil {
 		if !os.IsExist(err) {
 			return nil, fmt.Errorf("%w: %v", ErrFailedToInitialize, err)
 		}
--- a/internal/executor/cache/cache.go
+++ b/internal/executor/cache/cache.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 const bufSize = 10 * 1024 * 1024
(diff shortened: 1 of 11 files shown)


internal/executor/cache/putop.go
```diff
@@ -0,0 +1,34 @@
+package cache
+
+import (
+	"fmt"
+	"os"
+)
+
+type PutOperation struct {
+	tmpBlobFile   *os.File
```
fkorotkov, line 9: let's use a buffered writer here

internal/executor/cache/cache.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return &PutOperation{
+		tmpBlobFile:   tmpBlobFile,
+		finalBlobPath: c.blobPath(key),
+	}, nil
+}
+
+func (c *Cache) blobPath(key string) string {
+	// Sanitize user-controlled data by hashing it
```
fkorotkov, line 71: Might not be necessary since cache keys are always in this form `<cache name>-<sha256>`. Plus will be easier for users to find the blobs for manual inspection.

internal/executor/testdata/cache/.cirrus.yml
```diff
@@ -0,0 +1,17 @@
+container:
+  image: debian:latest
+
+fill_cache_task:
+  test_cache:
+    folder: cache
+    populate_script:
+      - mkdir cache
+      - touch cache/from-populate-script
```
fkorotkov, line 9: We can do better! Let's store a 100Mb file inside `cache`

```suggestion
      - cd cache
      - dd if=/dev/urandom of=from-populate-script bs=10M count=10
```

internal/executor/cache/putop.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/17/commits/424a4fa3523a55ea666e4909c0595b74893f5826.

internal/executor/cache/cache.go
edigaryev: It's technically possible for the arbitrary code that runs in the task's environment to impersonate the agent and connect to the CLI over new (or existing) RPC connection.

So this depends on whether we trust this environment or not.

edigaryev: Alternatively, we can validate the key using a simple regular expression.

internal/executor/cache/cache.go
```diff
@@ -9,6 +10,8 @@ import (
 	"path/filepath"
 )
 
+const bufSize = 10 * 1024 * 1024
```
fkorotkov, line 13: let's do 1Mb since agent streams in 1Mb chunks

```suggestion
const bufSize = 1024 * 1024 // 1Mb chunk size from the agent
```

internal/executor/cache/cache.go
fkorotkov: Yeah, can do something like:

```suggestion
	isSafeChar := func(c rune) bool { 
		if '0' <= c && c <= '9' { return true }
		if 'a' <= c && c <= 'z' { return true }
		if 'A' <= c && c <= 'Z' { return true }
		if c == '-' || c == '_' { return true }
		return false
	}
	isNotSafeChar := func(c rune) bool { 
		return !isSafeChar(c)
	}
	containsNonSafeChar := strings.IndexFunc(key, isNotSafeChar) > 0
	if !containsNonSafeChar {
		// only safe characters so no need for a sanitization
		return key
	}
	// Sanitize user-controlled data by hashing it
```

internal/executor/cache/cache.go
edigaryev: See 85c575b4d2975aff174a2e29b060424f34ced7cc.

internal/executor/cache/cache.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/17/commits/379f4a2983affb6a68db75cbf80e38034e0ccb89.

PR #19, opened 2020-08-03
Improved gRPC dial

fkorotkov:
* use retries
* block until connection is established
* bundle certificates via gocertifi

Similar to how the agent dials the rpc endpoint


edigaryev:
It seems that this needs to rebased onto `master` for the tests to pass.

PR #20, opened 2020-08-04
Integrate Echelon

fkorotkov:
cli-mark-2

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -26,6 +26,8 @@ github.com/cespare/xxhash v1.1.0/go.mod h1:XrSqR1VqqWfGrhpAt58auRo0WTKS1nRRg3ghf
 github.com/cirruslabs/cirrus-ci-agent v1.6.0 h1:Jvwjf16u+ZlomwmtSMI/4JijdJjrLgV9qrb8Anzarsg=
 github.com/cirruslabs/cirrus-ci-agent v1.6.0/go.mod h1:gx+B38qqOeK6MMlXYpKkuamrG/LxSpnhWx5/pMvUaAc=
 github.com/cirruslabs/cirrus-ci-annotations v0.0.0-20200716124710-e1adb4748203/go.mod h1:98qD7HLlBx5aNqWiCH80OTTqTTsbXT69wxnlnrnoL0E=
+github.com/cirruslabs/echelon v1.1.2 h1:715qznBkbAUyN8sYEngerDQzwpxaKE2pODhiRz5znqM=
+github.com/cirruslabs/echelon v1.1.2/go.mod h1:UhKEYRgma5rEMkrapp5VOfYln1aSoPSV/3BMd8sqG+o=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/containerd/containerd v1.3.6 h1:SMfcKoQyWhaRsYq7290ioC6XFcHDNcHvcEMjF6ORpac=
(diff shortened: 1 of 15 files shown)


internal/commands/run.go
```diff
@@ -43,6 +46,9 @@ func run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	if !verbose {
```
edigaryev, line 49: These two things (verbosity and interactivity) seem to be orthogonal: you can have both interactive and verbose output at the same time.

Take a look at `-1` argument of the [`ls`](https://man.openbsd.org/ls) command if you want an option to disable interactive output.

cmd/cirrus/main.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func main() {
-	// Set up signal interruptible context
+	// Set up signal interruptable context
```
edigaryev, line 12: https://en.wiktionary.org/wiki/interruptible

go.sum
```diff
@@ -22,6 +22,14 @@ github.com/cespare/xxhash v1.1.0/go.mod h1:XrSqR1VqqWfGrhpAt58auRo0WTKS1nRRg3ghf
...
 github.com/cirruslabs/cirrus-ci-annotations v0.0.0-20200708182643-6441beff996d/go.mod h1:98qD7HLlBx5aNqWiCH80OTTqTTsbXT69wxnlnrnoL0E=
+github.com/cirruslabs/echelon v0.0.0-20200803173629-85f41761ba61 h1:/oV6FLQLFQ21k2tgbkgV56D4jfCJAopDcObMos+Zibc=
+github.com/cirruslabs/echelon v0.0.0-20200803173629-85f41761ba61/go.mod h1:c57rTdFNE3E34PRYuWXFYZyb+xDozdqS5Fw5TlM7wxg=
+github.com/cirruslabs/echelon v0.0.0-20200803185240-d7b99d045349 h1:ffYD8ueB/pVjBAAIcpaG/0FR5jvAG0Viro09JQx8oQQ=
+github.com/cirruslabs/echelon v0.0.0-20200803185240-d7b99d045349/go.mod h1:c57rTdFNE3E34PRYuWXFYZyb+xDozdqS5Fw5TlM7wxg=
+github.com/cirruslabs/echelon v0.0.0-20200803202207-2e5d9027b8c4 h1:cVJsDiaL/0KFvurL+uzw6KX5zaVrvky2JRlXVqX2u2M=
+github.com/cirruslabs/echelon v0.0.0-20200803202207-2e5d9027b8c4/go.mod h1:c57rTdFNE3E34PRYuWXFYZyb+xDozdqS5Fw5TlM7wxg=
+github.com/cirruslabs/echelon v0.0.0-20200803204805-aa942972b98a h1:dX07a1HRRke/WQBgj3S13eOnGqZmmQSF/fE5m8jhxTA=
+github.com/cirruslabs/echelon v0.0.0-20200803204805-aa942972b98a/go.mod h1:c57rTdFNE3E34PRYuWXFYZyb+xDozdqS5Fw5TlM7wxg=
```
edigaryev, line 32: This clearly needs `go mod tidy`.

cmd/cirrus/main.go
fkorotkov: 🤷‍♂️ https://en.wiktionary.org/wiki/interruptable

IntelliJ's dictionary suggested it. Will revert. 😅

internal/commands/run.go
fkorotkov: I don't see how it can play well with the incremental redrawing. I can try to add log levels to echelon and have description only show messages for a particular log level. Then we can try to replace logrus completely with echelon. 🤔 What do you think?

edigaryev: That's an interesting idea. Could the echelon library provide a logrus-compatible interface to simplify integrations?

This way the only thing we'll have to do here is to swap the logger.

fkorotkov: You still need to create the hierarchy but the `Debugf` and context things can be replicated. Will experiments with it. 👍

fkorotkov:
Need to update Gif though. Will do it tomorrow.

fkorotkov:
This is now ready for another round of review. Now Echelon supports interactive mode like on the GIF as well as a simple colored output for dump terminals like in CI.

fkorotkov:
Hold on, will fix tests first. 😅

fkorotkov:
Fixed tests 👌😅

PR #21, opened 2020-08-04
A task pattern argument to only run specific tasks

edigaryev:
Some examples: `cirrus run "spell check"` or `cirrus run lint,test`.

I'm not sure if we actually want the task pattern to be comma-separated, because in the future other operators might be introduced (e.g: https://cirrus-ci.org/guide/writing-tasks/#supported-operators) yet, it is how it is.


fkorotkov:
There might be an issue when filter can filter out a dependency. I think in this scenario the other cask won't start executing at all. For, example:

```
container:
  image: alpine: latest

foo_task:
  script: true

bar_task:
  depends_on: foo
  script: true
```

If we pass "bar" as a filter it doesn't look it will start executing.

Also I think `Build` shouldn't know about filtration. Once tasks are parsed we can filter them out and also "fix" dependencies for remaining tasks.

We can avoid all of the above things by just not allowing a patter and literally run a single task by name for now.

edigaryev:
> We can avoid all of the above things by just not allowing a patter and literally run a single task by name for now.

Consider a `cirrus run primary` invocation with the following `.cirrus.yml`:

```yaml
container:
  image: debian:latest

secondary_task:
  script: true

primary_task:
  depends_on: secondary
  script: true
```

So even in the simplest case still need to implement a bit more complicated logic to resolve this automatically.

> Also I think `Build` shouldn't know about filtration. Once tasks are parsed we can filter them out and also "fix" dependencies for remaining tasks.

I'm not sure if that would help, since in the end the `Build` will still need to call the "dependency resolution engine" to choose the next task from a slice of tasks that you're proposing to pre-filter.

And if you're going to move that "dependency resolution engine" (that can also filter tasks) to a separate structure — it'd need to know about tasks and their states (which `Build` already knows about).

fkorotkov:
Maybe I wasn't clear about the filtering and fixing dependencies. I meant that once you parsed all tasks, you find the one matching the filter, find all dependencies of them recursively and only pass this subset of tasks to Buiild so it doesn't know about any filtration that happened before and simply needs to run all the tasks passed to it.

edigaryev:
> Maybe I wasn't clear about the filtering and fixing dependencies. I meant that once you parsed all tasks, you find the one matching the filter, find all dependencies of them recursively and only pass this subset of tasks to Buiild so it doesn't know about any filtration that happened before and simply needs to run all the tasks passed to it.

I see, so what's left to do here is to add a topological sort of the tasks (with filtering) before passing them to `Build`?

fkorotkov:
Yes, or only allow to run a single task without dependencies (you den clear dependencies of a task once you found it).

I'm not sure if it always an intent of a user to run a task **with dependencies** when running by name. 🤔

edigaryev:
> I'm not sure if it always an intent of a user to run a task **with dependencies** when running by name. 🤔

But on the other hand it was their intent to to add a `depends_on` directive into `.cirrus.yml`, right?

Also here's some interesting statistics from GitHub search: only [355](https://github.com/search?q=depends_on+filename%3A.cirrus.yml&type=Code) configurations out of [3,161](https://github.com/search?q=filename%3A.cirrus.yml&type=Code) (roughly 11%) use `depends_on`. The truth is, however, lies in what these dependent tasks actually do.

The other point is that automatically running the specified task's dependencies (with e.g. `cirrus run task`) makes behavior consistent with the `cirrus run`.

In the end we can always add a flag (e.g. `--no-deps`) when such need arises.

fkorotkov:
Or we can add `--with-deps` flag afterwards. IMO when I want to run a task, I want to run a single task. I see this as the most common scenario, a user sees that CI is failing due to some tests, the users checks out the same SHA, then types `cirrus run tests` and expects to see `tests` task running to reproduce the failure. Instead the user will see `lint` task running first instead the desired one.

edigaryev:
> Or we can add `--with-deps` flag afterwards. IMO when I want to run a task, I want to run a single task. I see this as the most common scenario, a user sees that CI is failing due to some tests, the users checks out the same SHA, then types `cirrus run tests` and expects to see `tests` task running to reproduce the failure. Instead the user will see `lint` task running first instead the desired one.

See https://github.com/cirruslabs/cirrus-cli/pull/21/commits/00eeeef74ccc81e91ef4ddf2a8761adc7b921101.

internal/executor/executor.go
```diff
@@ -36,6 +40,9 @@ func New(projectDir string, tasks []*api.Task, opts ...Option) (*Executor, error
 		e.logger.Out = ioutil.Discard
 	}
 
+	// Filter tasks (e.g. if a user wants to run only a specific task without dependencies)
+	tasks = e.taskFilter(tasks)
```
fkorotkov, line 44: I'm afraid it will break `GetNextTask` which can now return `nil` in this case when filtering for `bar` task:

```yaml
foo_task:
  script: echo 'Foo'

bar_task:
  depends_on: foo
  script: echo 'Bar'
```

edigaryev: Thanks, that's a pretty good catch!

https://github.com/cirruslabs/cirrus-cli/pull/21/commits/bf7f1bfe9e3f2f6ab5e0b49c3bda057979282a21

I've also thought about additionally preventing this in the `Build.taskHasUnresolvedDependencies()` itself, but I think that this will only be a burden with passing down the errors, since currently we're trusting the `.cirrus.yml` parser and in the future the parser will be completely in-house.

AlexWayfer:
As I understand, it doesn't work for me without explicit `name:`.

I have such `.cirrus.yaml` file: https://github.com/AlexWayfer/flame-cli/blob/0693eb3/.cirrus.yaml

There are `rspec_task`, as you can see, even without `matrix`.

```
> sudo cirrus run "rspec" -e CIRRUS_BRANCH=main -o simple
Error: task filter yielded no matches: none of the 0 task(s) were matched using a "rspec" filter
2022/09/12 12:58:34 task filter yielded no matches: none of the 0 task(s) were matched using a "rspec" filter
```

PR #26, opened 2020-08-11
Docker Pipe support

edigaryev:
https://cirrus-ci.org/guide/docker-pipe/


internal/executor/instance/pipe.go
```diff
@@ -0,0 +1,78 @@
...
+		params := &Params{
+			Image:             stage.Image,
+			CPU:               pi.CPU,
+			Memory:            pi.Memory,
+			CommandFrom:       stage.CommandFrom,
+			CommandTo:         stage.CommandTo,
+			WorkingVolumeName: workingVolume.Name(),
+		}
+
+		if err := RunDockerizedAgent(ctx, config, params); err != nil {
```
fkorotkov, line 68: What will happen here if the first stage fails? Seems there will be no error and a pipe will proceed, wouldn't it?

edigaryev: Fixed via https://github.com/cirruslabs/cirrus-cli/pull/26/commits/772c9e655f1d4600e40201ace81c88eac46ded9e.

fkorotkov:
I just realized that CLI never calls `ParseConfig` with `FilesContents` which prevents to have an accurate cache key and reuse an image built in the cloud.

PR #29, opened 2020-08-13
Support execution behavior on the CLI side

edigaryev:
This essentially introduces per-command status fields and lets us answer the question of e.g. whether a task containing `ON_FAILURE` command is completed, failed and whatnot.

### DIFF
--- a/internal/executor/build/command.go
+++ b/internal/executor/build/command.go
@@ -1,0 +1,31 @@
+package build
+
+import (
+	"github.com/cirruslabs/cirrus-ci-agent/api"
+	"github.com/cirruslabs/cirrus-cli/internal/executor/build/taskstatus"
+	"sync"
+)
+
+type Command struct {
+	status taskstatus.Status
+
+	// Original Protocol Buffers structure for reference
+	ProtoCommand *api.Command
+
+	// A mutex to guarantee safe accesses from both the main loop and gRPC server handlers
+	Mutex sync.RWMutex
+}
+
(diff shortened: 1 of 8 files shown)


internal/executor/build/command.go
```diff
@@ -0,0 +1,31 @@
+package build
+
+import (
+	"github.com/cirruslabs/cirrus-ci-agent/api"
+	"github.com/cirruslabs/cirrus-cli/internal/executor/build/taskstatus"
+	"sync"
+)
+
+type Command struct {
+	status taskstatus.Status
```
fkorotkov, line 10: `taskstatus` looks a bit weird here. It can also be of `New` value. 🤔 What do you think about introducing `commandstatus`?

edigaryev: >It can also be of `New` value.

What's wrong with that?

fkorotkov: `New` just sounds a bit weird. See how it's [defined in GraphQL](https://github.com/cirruslabs/cirrus-ci-web/blob/master/schema.graphql#L644-L651). I think for consistency we should also rename New to [Created](https://github.com/cirruslabs/cirrus-ci-web/blob/master/schema.graphql#L691) in a separate PR.

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/29/commits/1ca413f6d7ae530b7f92d9caabda2d2bbd26a661.

PR #33, opened 2020-08-16
Support Dockerfile as CI environment

### LINKED ISSUE #32: Support Dockerfile as CI environment
https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment

edigaryev:
Resolves #32.

### DIFF
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -63,6 +63,24 @@ func NewFromProto(instance *api.Task_Instance, commands []*api.Command) (Instanc
 			Memory: pipe.Memory,
 			Stages: stages,
 		}, nil
+	case "prebuilt_image":
+		var prebuilt api.PrebuiltImageInstance
+		if err := proto.Unmarshal(instance.Payload, &prebuilt); err != nil {
+			return nil, err
+		}
+
+		// PrebuiltImageInstance is currently missing the domain part to craft the full image name
+		// used in the follow-up tasks.
+		//
+		// However, since currently the only possible value is "gcr.io",
+		// we simply craft the image name manually using that hardcoded value.
+		image := path.Join("gcr.io", prebuilt.Repository) + ":" + prebuilt.Reference
+
+		return &PrebuiltInstance{
+			Image:      image,
+			Dockerfile: prebuilt.DockerfilePath,
+			Arguments:  prebuilt.Arguments,
+		}, nil
 	default:
 		return nil, fmt.Errorf("%w: %s", ErrUnsupportedInstance, instance.Type)
 	}
@@ -94,14 +112,23 @@ func RunDockerizedAgent(ctx context.Context, config *RunConfig, params *Params) 
 		return err
 	}
 
-	logger.Debugf("pulling image %s", params.Image)
-	progress, err := cli.ImagePull(ctx, params.Image, types.ImagePullOptions{})
-	if err != nil {
-		return err
+	var noPull bool
+
+	// Handle prebuilt instance images being available only locally
+	if strings.HasPrefix(params.Image, "gcr.io/cirrus-ci-community/") {
+		noPull = true
 	}
-	_, err = io.Copy(ioutil.Discard, progress)
-	if err != nil {
-		return err
+
+	if !noPull {
+		logger.Debugf("pulling image %s", params.Image)
+		progress, err := cli.ImagePull(ctx, params.Image, types.ImagePullOptions{})
+		if err != nil {
+			return err
+		}
+		_, err = io.Copy(ioutil.Discard, progress)
+		if err != nil {
+			return err
+		}
 	}
 
 	logger.Debugf("creating container using working volume %s", params.WorkingVolumeName)
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -64,7 +64,7 @@ func NewFromProto(instance *api.Task_Instance, commands []*api.Command) (Instanc
 			Stages: stages,
 		}, nil
 	default:
-		return nil, ErrUnsupportedInstance
+		return nil, fmt.Errorf("%w: %s", ErrUnsupportedInstance, instance.Type)
 	}
 }
 
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -112,14 +112,19 @@ func RunDockerizedAgent(ctx context.Context, config *RunConfig, params *Params) 
 		return err
 	}
 
-	var noPull bool
+	// Check if the image is missing and pull it if needed
+	var needToPull bool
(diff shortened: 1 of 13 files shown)


internal/executor/instance/prebuilt.go
```diff
@@ -0,0 +1,177 @@
...
+	// Create an archive with the build context
+	archivePath, err := CreateTempArchive(config.ProjectDir)
+	if err != nil {
+		return err
+	}
+	file, err := os.Open(archivePath)
+	if err != nil {
+		return err
+	}
+	if err := os.Remove(archivePath); err != nil {
```
fkorotkov, line 111: Why is it removed right away?

internal/executor/instance/prebuilt.go
```diff
@@ -0,0 +1,177 @@
...
+	"strings"
+)
+
+type PrebuiltInstance struct {
+	Image      string
+	Dockerfile string
+	Arguments  map[string]string
+}
+
+func CreateTempArchive(dir string) (string, error) {
```
fkorotkov, line 24: We should look into reusing parts of `internal/targz/targz.go` from the agent here.

internal/executor/instance/instance.go
```diff
@@ -98,14 +116,23 @@ func RunDockerizedAgent(ctx context.Context, config *RunConfig, params *Params)
...
+	if strings.HasPrefix(params.Image, "gcr.io/cirrus-ci-community/") {
+		noPull = true
 	}
-	_, err = io.Copy(ioutil.Discard, progress)
-	if err != nil {
-		return err
+
+	if !noPull {
+		logger.WithContext(ctx).Debugf("pulling image %s", params.Image)
+		progress, err := cli.ImagePull(ctx, params.Image, types.ImagePullOptions{})
```
fkorotkov, line 128: Maybe we should just check local cache if the image exists and pull only if not? Might be useful for folks to check their custom build environments locally. I some use cases where people crafted an image for CI manually or locally and just pushed it to Docker hub. With the current pull there is no way for them to test new images without pushing them. 🤔

internal/executor/instance/prebuilt.go
edigaryev: `targz` name suggests that it wouldn't be compatible with Docker Engine API right away, since the latter expects a simple tar archive, so it would require some changes. Moreover, the implementation seems to be doing strange things like zeroing the modification timestamps, which is crucial for the correct operation of many things (`make`, etc).

Perhaps we should keep it as is since it's not that huge and does what it's designed to do, and not multiple things at once?

internal/executor/instance/prebuilt.go
edigaryev: Because you can remove the file right away on Unix-like OS'es and keep working with it. Why would we wait, if it wouldn't be needed in the future anyway? But I'm not sure about Windows yet.

fkorotkov: But wouldn't Docker Client work with it and send it as context? This Unix behaviour is non intuitive that you can open a file, remove it and still work with it. What will the `err` be? How it can be sure file will be removed. Pretty confusing IMO 😅 Maybe just switch to `defer os.Remove(archivePath)`?

internal/executor/instance/instance.go
edigaryev: >Maybe we should just check local cache if the image exists and pull only if not?

>With the current pull there is no way for them to test new images without pushing them.

This is a valid use-case but note that this would prevent the image from ever being updated, which makes some things harder to diagnose when working with rapidly-changing images.

The other option is to always pull the image but never bail-out due to a failed pull, just simply log the warning message and continue.

For now the solution you've proposed indeed seems to be adequate, but in the future we'd probably have to add a new command-line argument for enabling the pulls since a configuration file may contain a lot of images, and manually updating them would be too tedious.

internal/executor/instance/prebuilt.go
fkorotkov: I was thinking of refactoring `targz` in two methods. Something like https://github.com/cirruslabs/cirrus-ci-agent/pull/39

internal/executor/instance/prebuilt.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/33/commits/056dbde64713a3e6a033faf06693f4a749608d66.

internal/executor/instance/prebuilt.go
edigaryev: What about the other issue?

internal/executor/instance/instance.go
fkorotkov: `docker run` does use a local image if available instead of always trying to pull it. Behaviour I described is closer to this. We can think of having a `cirrus pull` command that can re-pull images needed by tasks. 🤔

internal/executor/instance/prebuilt.go
edigaryev: Seems that we've decided to keep this as is, at least for now.

internal/executor/instance/instance.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/33/commits/92e0fc87b200f32626ba1ef0ac9b40388f4d9c56.

edigaryev:
Also need to handle this here: https://github.com/cirruslabs/cirrus-cli/pull/26#issuecomment-674897151.

edigaryev:
> Also need to handle this here: [#26 (comment)](https://github.com/cirruslabs/cirrus-cli/pull/26#issuecomment-674897151).

Implemented in https://github.com/cirruslabs/cirrus-cli/pull/33/commits/f65fd62cdef696a68dee592ce03641f0527ea4c6.

pkg/parser/parser.go
```diff
@@ -109,5 +120,35 @@ func (p *Parser) ParseFromFile(path string) (*Result, error) {
...
+		contents, err := ioutil.ReadFile(filepath.Join(filepath.Dir(path), prebuilt.Dockerfile))
+		if err != nil {
+			return nil, fmt.Errorf("%w: %v", ErrFilesContents, err)
+		}
+		filesContents[prebuilt.Dockerfile] = string(contents)
+	}
+
+	// Parse again with the file contents supplied
+	p.FilesContents = filesContents
 	return p.Parse(string(config))
```
fkorotkov, line 153: You can skip it `filesContents` is empty and return previously parsed tasks.

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/33/commits/8f96b9ed4f1592eb7677b6f5a4c6e8e186258e20.

PR #34, opened 2020-08-17
Dirty mode for running tasks

### LINKED ISSUE #27: Dirty mode for running tasks
Right now `run` command `rsync`s working directory to a container respecting `.gitignore`. It takes time and might not be always necessary because some tasks like linting are readonly. It will be good to have a `--dirty` flag so instead of `rsync`ing project folder will be mounted directly.

edigaryev:
Resolves #27.

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -15,6 +15,7 @@ import (
 
 var ErrRun = errors.New("run failed")
 
+var dirty bool
 var environment []string
 var runFile string
 var verbose bool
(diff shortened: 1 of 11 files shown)


internal/commands/run.go
```diff
@@ -105,6 +111,7 @@ func newRunCmd() *cobra.Command {
 		Args:  cobra.MaximumNArgs(1),
 	}
 
+	cmd.PersistentFlags().BoolVar(&dirty, "dirty", false, "mount project directory in read-write mode")
```
fkorotkov, line 114: Maybe add description of both cases:

```suggestion
	cmd.PersistentFlags().BoolVar(&dirty, "dirty", false, "if true mounts project directory in read-write mode. otherwise copies files respecting .gitignore")
```

edigaryev, line 114: There's no "true" or "false" on the UI side, it's a flag: you either set it or not.

```suggestion
	cmd.PersistentFlags().BoolVar(&dirty, "dirty", false, "if set the project directory will be mounted in read-write mode, otherwise the project directory files are copied, taking .gitignore into account")
```

What do you think about this instead?

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/34/files#r471635981.

fkorotkov: LGTM

PR #39, opened 2020-08-19
Stable order of task execution

fkorotkov:
So the order of execution is always the same

### DIFF
--- a/internal/executor/build/build.go
+++ b/internal/executor/build/build.go
@@ -94,5 +94,6 @@ func (b *Build) GetNextTask() (result *Task) {
 			result = task
 		}
 	}
+
 	return
 }
--- a/internal/executor/build/build.go
+++ b/internal/executor/build/build.go
@@ -6,7 +6,6 @@ import (
 	"github.com/cirruslabs/cirrus-cli/internal/executor/cache"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"math"
 	"path/filepath"
 )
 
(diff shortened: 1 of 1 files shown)


edigaryev:
Perhaps you could only use a single variable to do the job right?

```diff
diff --git a/internal/executor/build/build.go b/internal/executor/build/build.go
index e6741cc..552d320 100644
--- a/internal/executor/build/build.go
+++ b/internal/executor/build/build.go
@@ -84,16 +84,20 @@ func (b *Build) taskHasUnresolvedDependencies(task *Task) bool {
        return false
 }
 
-func (b *Build) GetNextTask() *Task {
+func (b *Build) GetNextTask() (result *Task) {
        for _, task := range b.tasks {
                if b.taskHasUnresolvedDependencies(task) {
                        continue
                }
 
-               if task.Status() == taskstatus.New {
-                       return task
+               if task.Status() != taskstatus.New {
+                       continue
+               }
+
+               if result == nil || task.ID < result.ID {
+                       result = task
                }
        }
 
-       return nil
+       return
 }
```

internal/executor/build/build.go
```diff
@@ -84,16 +84,15 @@ func (b *Build) taskHasUnresolvedDependencies(task *Task) bool {
...
 
-		if task.Status() == taskstatus.New {
-			return task
+		if result == nil || task.ID < result.ID {
+			result = task
 		}
 	}
-
-	return nil
+	return
```
edigaryev, line 97: It would be nice to make this consistent with the rest of the code in this file by adding an empty line before `return`.

edigaryev, line 97:
```suggestion

	return
```

edigaryev (approved):

PR #42, opened 2020-08-20
Cache gRPC connections

fkorotkov:
There are occasional slowness in tests while dialing gRPC endpoint. My theory is that tests are running in parallel and creation of a new connection for every parsing causing the issue since usually you have a persistent connection for your app and don't create one for each RPC.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -15,6 +15,7 @@ import (
 	"google.golang.org/grpc/status"
 	"io/ioutil"
 	"path/filepath"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


edigaryev:
It would be nice if this PR would only solve a single problem (without including things like function rearrangement), so that it would be easily identifiable and reversible in the future.

fkorotkov:
Sorry about that. Check now.

edigaryev:
It seems that the connections might linger in the `clientsCache` map for longer that it's required when running tasks via a CLI.

And looks like the caching trick didn't work as expected (https://cirrus-ci.com/task/6333367975149568) and we should probably investigate a proper way to do gRPC retries, which, compared to caching, will be beneficial both in tests and in the production usage.

fkorotkov:
Yeah,

PR #47, opened 2020-08-25
Proof of concept of syntax sugar

fkorotkov:
Essentially this is how we can create helper function for creating Cirrus CI tasks


pkg/larker/testdata/simple-task/expected.yaml
```diff
@@ -0,0 +1,5 @@
+task:
+  container:
+    image: debian:latest
+  script:
+    - printenv
```
edigaryev, line 5: A newline is missing here (there are other occurrences too, marked with a red "no entry" icon).

pkg/larker/larker.go
```diff
@@ -26,6 +27,9 @@ func New(opts ...Option) *Larker {
 		fs: fs.NewDummyFileSystem(),
 	}
 
+	// weird global init by Starlark
```
edigaryev, line 30: Perhaps a better idea would be to explain why do we enable floats here instead.

pkg/larker/larker_test.go
```diff
@@ -3,22 +3,28 @@ package larker_test
...
 
-// TestSimpleTask ensures that .cirrus.star is able to generate the simplest possible configuration
-// that gets executed without any problems.
+// TestSimpleTask ensures that .cirrus.star is able to generate the simplest possible configuration.
 func TestSimpleTask(t *testing.T) {
-	dir := testutil.TempDirPopulatedWith(t, "testdata/simple-task")
+	validateExpected(t, "testdata/simple-task")
+}
+
+// TestSugarCoatedTask ensures that .cirrus.star is able to use global functions for task generation.
```
edigaryev, line 21:
```suggestion
// TestSugarCoatedTask ensures that .cirrus.star is able to use imported functions for task generation.
```

edigaryev (changes requested):

edigaryev (approved):

fkorotkov:
BTW I found an issue with the converter yesterday and finally found out the reason: `startlark.Int` is not just an `int`. Fixed in 2b79757cd130a5490b49229856f82b89df014993

edigaryev (approved):

PR #49, opened 2020-08-28
Don't include version in published binary

fkorotkov:
So the CLI can be installed like this:

```bash
curl -o cirrus https://github.com/cirruslabs/cirrus-cli/releases/latest/download/cirrus-$(uname | tr '[:upper:]' '[:lower:]')-amd64 \
  && sudo mv cirrus /usr/local/bin/cirrus && sudo chmod +x /usr/local/bin/cirrus
```


.gitignore
```diff
@@ -1,15 +1,2 @@
...
-# Test binary, built with `go test -c`
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
-# Dependency directories (remove the comment below to include it)
-# vendor/
+.idea
+dist
```
edigaryev, line 2: Missing newline:

```suggestion
dist

```

.gitignore
```diff
@@ -1,15 +1,2 @@
...
-
-# Test binary, built with `go test -c`
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
-# Dependency directories (remove the comment below to include it)
-# vendor/
+.idea
```
edigaryev, line 1: Let's comment this in similar way it's been done before and add a slash to signify that `.idea` is a directory:

```suggestion
# IntelliJ IDEA
.idea/
```

.gitignore
edigaryev, line 2:
```suggestion

# GoReleaser
dist
```

edigaryev (approved):

PR #54, opened 2020-08-31
Reworked documentation

fkorotkov:
Added a separate installation guide, added an example and some other stuff. @edigaryev what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,62 +3,91 @@
 [![Build Status](https://api.cirrus-ci.com/github/cirruslabs/cirrus-cli.svg?branch=master)](https://cirrus-ci.com/github/cirruslabs/cirrus-cli)
 
 Cirrus CLI is a tool for running containerized tasks reproducibly in any environment. Run your tasks locally the same way
-they are executed in CI or on your colleague's computer. Immutability of Docker containers ensures your project will compile
-years form now regarding what versions of packages you have locally.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,62 +3,93 @@
...
-go get github.com/cirruslabs/cirrus-cli/...
-```
-
-This will build and place the `cirrus` binary in `$GOPATH/bin`.
-
-To be able to run `cirrus` command from anywhere, make sure the `$GOPATH/bin` directory is added to your `PATH`
-environment variable (see [article in the Go wiki](https://github.com/golang/go/wiki/SettingGOPATH) for more details).
+* [Prebuilt Binary](intallation.md#prebuilt-binary)
+* [Golang](intallation.md#golang)
+* [Github Actions](intallation.md#github-actions)
```
edigaryev, line 15: When looking at this list rendered in Markdown, it feels like these four CI-related options should be grouped together to separate them from the rest of the "standard" options at first glance.

What do you think?

intallation.md
```diff
@@ -0,0 +1,101 @@
+# Index
```
edigaryev, line 1: Besides the typo, how about we call this `INSTALL.md`, so that it would be consistent with the `README.md` and more easily discoverable in the repository's root?

fkorotkov: `Index` it's like index in a book. Didn't mean `Install`. 😅

edigaryev (approved):

Issue #61, opened 2020-09-06
`cirrus run` fails trying to open a port on a nonexistent IP


jaqx0r:
Using the cirrus config from github.com/google/mtail/.cirrus.yml and a freshly "go get" cirrus cli:

```
$ cirrus run
panic: listen tcp 172.17.0.1:0: bind: cannot assign requested address

goroutine 1 [running]:
github.com/cirruslabs/cirrus-cli/internal/executor/rpc.(*RPC).Start(0xc00009e060)
	/home/jaq/go/pkg/mod/github.com/cirruslabs/cirrus-cli@v0.5.0/internal/executor/rpc/rpc.go:119 +0x359
github.com/cirruslabs/cirrus-cli/internal/executor.(*Executor).Run(0xc0001ed2f0, 0xf4e920, 0xc0003d8300, 0x0, 0x0)
	/home/jaq/go/pkg/mod/github.com/cirruslabs/cirrus-cli@v0.5.0/internal/executor/executor.go:82 +0x62
github.com/cirruslabs/cirrus-cli/internal/commands.run(0xc0003ccb00, 0x1582820, 0x0, 0x0, 0x0, 0x0)
	/home/jaq/go/pkg/mod/github.com/cirruslabs/cirrus-cli@v0.5.0/internal/commands/run.go:158 +0x5f5
github.com/spf13/cobra.(*Command).execute(0xc0003ccb00, 0x1582820, 0x0, 0x0, 0xc0003ccb00, 0x1582820)
	/home/jaq/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003cc580, 0xe410d8, 0xc00029b440, 0xc0003dc6a0)
	/home/jaq/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
	/home/jaq/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
github.com/spf13/cobra.(*Command).ExecuteContext(...)
	/home/jaq/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:880
main.main()
	/home/jaq/go/pkg/mod/github.com/cirruslabs/cirrus-cli@v0.5.0/cmd/cirrus/main.go:27 +0x137
$ echo $?
2
```

[...]

PR #62, opened 2020-09-06
Include unique labels in logs

fkorotkov:
During testing #61 I noticed CLI printed bunch of `matrix_test` tasks without labels which made it impossible to identify tasks. Let's include unique labels in an echelon scope.

mtail

### DIFF
--- a/internal/executor/rpc/rpc.go
+++ b/internal/executor/rpc/rpc.go
@@ -178,7 +178,7 @@ func (r *RPC) ReportSingleCommand(
 		return nil, err
 	}
 
-	commandLogger := r.logger.Scoped(task.Name).Scoped(req.CommandName)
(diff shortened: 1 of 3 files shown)


internal/executor/rpc/rpc.go
```diff
@@ -116,7 +116,7 @@ func (r *RPC) Start() {
 
 	listener, err := net.Listen("tcp", address)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("failed to start RPC service on %s:0 with '%v'. do ", host, err))
```
edigaryev, line 119:
```suggestion
		panic(fmt.Sprintf("failed to start RPC service on %s: %v", address, err))
```

edigaryev (approved):

edigaryev (approved):

PR #68, opened 2020-09-07
Fixed Homebrew release

fkorotkov:
Goreleaser can only create formulas for archives but we only publish binaries. See https://github.com/goreleaser/goreleaser/issues/1103


fkorotkov:
Tested locally via `goreleaser --snapshot --skip-publish --rm-dist`

edigaryev (approved):

PR #69, opened 2020-09-08
Parser JSON tests

edigaryev:
This makes sure that we compare the actual parsing results (and not merely check the absence of errors) with the expected fixture contents.

`cirrus validate --json` can be used to generate the fixtures.

### DIFF
--- a/internal/testutil/api.go
+++ b/internal/testutil/api.go
@@ -22,7 +22,7 @@ func TasksToJSON(t *testing.T, tasks []*api.Task) []byte {
 			t.Fatal(err)
 		}
 
-		unmarshalledTasks = append(unmarshalledTasks, task)
+		unmarshalledTasks = append(unmarshalledTasks, unmarshalledTask)
 	}
 
 	res, err := json.MarshalIndent(unmarshalledTasks, "", "  ")
--- a/internal/testutil/api.go
+++ b/internal/testutil/api.go
@@ -1,0 +1,34 @@
+package testutil
+
+import (
+	"encoding/json"
+	"github.com/cirruslabs/cirrus-ci-agent/api"
+	"google.golang.org/protobuf/encoding/protojson"
+	"testing"
+)
+
+func TasksToJSON(t *testing.T, tasks []*api.Task) []byte {
+	var unmarshalledTasks []interface{}
+
+	for _, task := range tasks {
+		var unmarshalledTask interface{}
+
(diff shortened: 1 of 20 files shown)


internal/testutil/api.go
```diff
@@ -0,0 +1,34 @@
...
+	"testing"
+)
+
+func TasksToJSON(t *testing.T, tasks []*api.Task) []byte {
+	var unmarshalledTasks []interface{}
+
+	for _, task := range tasks {
+		var unmarshalledTask interface{}
+
+		marshalled, err := protojson.MarshalOptions{Indent: "  "}.Marshal(task)
```
fkorotkov, line 16: I wonder if we need `EmitUnpopulated: true` 🤔

fkorotkov (approved):

fkorotkov:
My bad. I only reviewed one commit. GitHub notification redirected me to the latest commit

internal/testutil/api.go
edigaryev: I think it will just tremendously increase the verbosity with the little benefit of detecting `.proto` changes, and these in turn would require us to re-generate the fixture each time a field which is not beneficial for the test case is added or removed.

pkg/parser/node/accessor.go
```diff
@@ -68,3 +62,24 @@ func (node *Node) GetSliceOfNonEmptyStrings() ([]string, error) {
...
 	}
 }
+
+func expandEnvironmentVariables(s string, env map[string]string) string {
+	const maxExpansionIterations = 10
+
+	for i := 0; i < maxExpansionIterations; i++ {
+		beforeExpansion := s
+
+		for key, value := range env {
```
fkorotkov, line 72: This is not deterministic and can lead to weird cases when you have two environment varaibales where one is a prefix of another. For example, `CIRRUS=true` and `CIRRUS_BRANCH=main`. Cirrus cloud sorts the environment variables from longest tot shortest before evaluating.

fkorotkov (commented):
Awesome! Left a small nit pick.

pkg/parser/node/accessor.go
edigaryev: Thanks for the nit, fixed in https://github.com/cirruslabs/cirrus-cli/pull/69/commits/486ef08a80f0747532d9588433786011b9e95d19.

fkorotkov (approved):

PR #75, opened 2020-09-10
Clamp CPU and memory to resources available for Docker daemon

edigaryev:
Currently both native and RPC YAML parsers default to 2 CPUs in case nothing else is specified, and on some VMs there is only 1 CPU.

Do our best effort and try to run the task anyway, since the user has already signified his intent to do so.

### DIFF
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -14,7 +14,9 @@ import (
 	"github.com/golang/protobuf/ptypes/any"
 	"io"
 	"io/ioutil"
+	"math"
 	"path"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -155,7 +157,7 @@ func RunDockerizedAgent(ctx context.Context, config *RunConfig, params *Params) 
 			},
 		},
 		Resources: container.Resources{
-			NanoCPUs: int64(params.CPU * nano),
+			NanoCPUs: int64(clampCPU(params.CPU) * nano),
 			Memory:   int64(params.Memory * mebi),
 		},
 	}
@@ -263,7 +265,7 @@ func runAdditionalContainer(
 	}
 	hostConfig := container.HostConfig{
 		Resources: container.Resources{
-			NanoCPUs: int64(additionalContainer.Cpu * nano),
+			NanoCPUs: int64(clampCPU(additionalContainer.Cpu) * nano),
 			Memory:   int64(additionalContainer.Memory * mebi),
 		},
 		NetworkMode: container.NetworkMode(fmt.Sprintf("container:%s", connectToContainer)),
@@ -313,4 +315,8 @@ func envMapToSlice(envMap map[string]string) (envSlice []string) {
 	}
 
 	return
+}
+
+func clampCPU(requested float32) float32 {
+	return float32(math.Min(float64(runtime.NumCPU()), float64(requested)))
 }
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -16,7 +16,6 @@ import (
 	"io/ioutil"
 	"math"
 	"path"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


fkorotkov:
In this case we should probably check memory as well...

What do you think about failing in this case but also adding a `--strict-resources=false` which will simply not set any resources for the containers?

edigaryev:
> What do you think about failing in this case but also adding a `--strict-resources=false` which will simply not set any resources for the containers?

I'm somewhat reluctant to add flags for behaviors that can be automatically sorted out without any user input since this will result in the flag creep in the future (try out the `docker run python:latest sh -c 'pip install uwsgi && uwsgi --help'` to get a feel of what that means).

Perhaps we can retrieve memory information in OS-independent way from `docker info` since we're already use Docker in this context.

fkorotkov:
`docker info` sounds good since you can also set the amount of CPUs and Memory available for Docker which is not the same as the OS resources.

edigaryev:
> `docker info` sounds good since you can also set the amount of CPUs and Memory available for Docker which is not the same as the OS resources.

See https://github.com/cirruslabs/cirrus-cli/pull/75/commits/2bc2f547e10ad03a7662eea6cc0f173dab44ebf9.

internal/executor/instance/instance.go
```diff
@@ -105,6 +104,21 @@ func RunDockerizedAgent(ctx context.Context, config *RunConfig, params *Params)
...
+	info, err := cli.Info(ctx)
+	if err != nil {
+		return err
+	}
+	availableCPU := float32(info.NCPU)
+	availableMemory := uint32(info.MemTotal / mebi)
+
+	params.CPU = clampCPU(params.CPU, availableCPU)
+	params.Memory = clampMemory(params.Memory, availableMemory)
+	for _, additionalContainer := range params.AdditionalContainers {
```
fkorotkov, line 117: I think in case of additional containers you can still run out of resources. A solution might be to get overall requested CPUs and Memory numbers, see if any of them are higher then the available resources and then scale all requested CPUs and memory proportionally.

Or if the requested resources are higher simply not set any limits. 🤔

edigaryev: I don't think these resource limits work similarly to Kubernetes, i.e. you can keep creating new containers as long as their individual limits fly below the radar.

fkorotkov: Interesting 🤷‍♂️

edigaryev:
```
# docker info | grep -E '(CPU|Memory)'
 CPUs: 1
 Total Memory: 987.2MiB
```

```
# for i in `seq 1 20`; do docker run -d --cpus 1 -m 1024M debian:latest sleep infinity;done
9cf924a37a5749db7f897d129104ea2891434b38ceeb04b41300a1876ba9578d
1fb3a76ac83aa95b5a4128cb5b40b603c22253f51c438a3c29172369efc7a980
03fd2c5c2ea068a5216b8c656ddcdfb984b622521b479d1b7e521eace0947b70
fd552e1762e622a299c8885da10d0d73bac5ee20cba20088870fbfc62f842e98
5ee2e74dd9f6f9d5c4d6fc9743c0666e05d22ce27bf7e5d20b570ae698a8a6cb
38694149c4814d78f01f7565875e860fea8cc2d694f0153ef5e8d19dedef01af
79f951266bb1515382ef7b213aafaa6ec108fdc7656fa53ce8dad15342986ce8
8a2ca06a36e2d148700ffcc417828d59ac579efb132c9ee48f1e4b51a8ba4507
483f4b664adf4ed6c66b5b43972fceed25021b50d596fdbe3e51c6ef594b1824
24a05ed8c9781ba4d17c83a9a29d72430f4edbe522c8fbafea60d56f9ed88a0d
14beb13e75a073b119c67b6512523e9886919a859c202d9edaa876d0aa4519d1
00150a8c0a85dcfd814e4a44f49b901edb9524f0445f1801e367bbf1646a4ede
7a432ad51b62c9491c27f8a577ca4858af9c81c2962db486d8462b41c46a99d7
e25541d83dbf7b280e97504757a32547b3833215abc78a300361a5f40b8ef5bf
116d14f86ed9b70dd9489041ae333ffcbc5193bd052609ee32a7faed124d7c6d
1fbcb5ee2db34cdb6371b5b5777765e3dd92959ebe122d35ba4b6743c9923bc8
48bdc66c5618307d6048f1938aaee2e98a96df2c4f334fd44e108eac17b44983
54e8dbee92595474be75d5e4a13ee9c2c204dc515e253ca45d68775624f2ba94
898c89e0825919e2a202e5a94889803bab8530611279aba9e679b1559316f3bb
3332830ceabff67f733d3b57adff93d16ecc6e666490269652579bef6fe1f74d
```

PR #76, opened 2020-09-10
Unique task description

fkorotkov:
There is a problem sometimes to distinguish task's output from an instruction output in dumb terminal mode.

```yaml
test_task:
  test_script: ./.ci/test.sh
```

The above config will produce a bit confusing output with two identical "Started 'test'". This change will fix it by making tasks scopes more descriptive.

### DIFF
--- a/internal/executor/build/task.go
+++ b/internal/executor/build/task.go
@@ -89,11 +89,11 @@ func (task *Task) ProtoCommands() []*api.Command {
 	return result
 }
 
-func (task *Task) UniqueName() string {
+func (task *Task) UniqueDescription() string {
 	if len(task.Labels) == 0 {
-		return task.Name
+		return fmt.Sprintf("%s Task", task.Name)
 	}
-	return task.Name + " " + strings.Join(task.Labels, " ")
+	return fmt.Sprintf("%s Task (%s)", task.Name, strings.Join(task.Labels, " "))
 }
 
 func (task *Task) FailedAtLeastOnce() bool {
--- a/internal/executor/build/task.go
+++ b/internal/executor/build/task.go
@@ -91,9 +91,9 @@ func (task *Task) ProtoCommands() []*api.Command {
 
 func (task *Task) UniqueDescription() string {
 	if len(task.Labels) == 0 {
-		return fmt.Sprintf("%s Task", task.Name)
+		return fmt.Sprintf("'%s' Task", task.Name)
 	}
-	return fmt.Sprintf("%s Task (%s)", task.Name, strings.Join(task.Labels, " "))
+	return fmt.Sprintf("'%s' Task (%s)", task.Name, strings.Join(task.Labels, " "))
 }
 
 func (task *Task) FailedAtLeastOnce() bool {
--- a/internal/executor/build/task.go
+++ b/internal/executor/build/task.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
+	"unicode/utf8"
 )
 
 const defaultTaskTimeout = 60 * time.Minute
@@ -90,10 +92,16 @@ func (task *Task) ProtoCommands() []*api.Command {
 }
 
 func (task *Task) UniqueDescription() string {
-	if len(task.Labels) == 0 {
-		return fmt.Sprintf("'%s' Task", task.Name)
+	name := task.Name
+	taskMessagePart := "task"
+	firstRune, _ := utf8.DecodeRuneInString(name)
+	if firstRune != utf8.RuneError && unicode.IsUpper(firstRune) {
+		taskMessagePart = "Task"
 	}
-	return fmt.Sprintf("'%s' Task (%s)", task.Name, strings.Join(task.Labels, " "))
+	if len(task.Labels) == 0 {
+		return fmt.Sprintf("'%s' %s", name, taskMessagePart)
+	}
+	return fmt.Sprintf("'%s' %s (%s)", name, taskMessagePart, strings.Join(task.Labels, " "))
 }
 
 func (task *Task) FailedAtLeastOnce() bool {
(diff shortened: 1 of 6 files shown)


edigaryev:
I'm not sure if augmenting the task scopes will be enough. Here's a simple configuration to demonstrate the problem:

```yaml
container:
  image: debian:latest

a_task:
  a_script: true
```

And the output of the `cirrus run -v` built from this branch:

```
gRPC server is listening at http://host.docker.internal:54396
running task a (0)
Started 'a Task'
Started 'Preparing execution environment...'
Preparing volume to work with...
'Preparing execution environment...' succeeded in 5.3s!
creating Docker client
creating container using working volume 9cc51baea854aa94a3b458f12b67f8048aa3e36261f450d17463b0c7c10affa6
starting container 399afd3865933b91947f97130204d281f4b35ea23ba56bc2c39f9f6ae9c80024
waiting for container 399afd3865933b91947f97130204d281f4b35ea23ba56bc2c39f9f6ae9c80024 to finish
received heartbeat
received heartbeat
Started 'a'
begin streaming logs
agent signal: child exited
received log chunk of 5 bytes
true

command succeeded
'a' succeeded in 1.0s!
container exited with <nil> error and exit code 0
cleaning up container 399afd3865933b91947f97130204d281f4b35ea23ba56bc2c39f9f6ae9c80024
task a (0) succeeded
'a Task' succeeded in 6.8s!
```

In this line:

```
Started 'a'
```

It's not quite evident by just looking at it what the `a` represents. Is it a task or a command?

fkorotkov:
There is `'a Task'` vs just `'a'`. But yeah, it not that evident here. But in case of GitHub Actions and Travis CI there are special renderers that will do folding. For regular console there is the interactive renderer that has a tree-like structure to figure it out.  I was thinking about making `'Instruction a'` scope instead of `'a'` but I think it's a bit too much for the interactive rendered. What do you think? I don't know how to make it generic and beautiful for all the different outputs. 😪

fkorotkov:
@edigaryev finally fixed all the linting issues 😅  PTAL

internal/executor/build/task.go
```diff
@@ -89,11 +89,11 @@ func (task *Task) ProtoCommands() []*api.Command {
 	return result
 }
 
-func (task *Task) UniqueName() string {
+func (task *Task) UniqueDescription() string {
 	if len(task.Labels) == 0 {
-		return task.Name
+		return fmt.Sprintf("'%s' Task", task.Name)
```
edigaryev, line 94: Perhaps the `Task` here and below should really be `task` to be consistent with the script type names?

fkorotkov: `'Tests (Go 1.15) Task` looks better then `'Tests (Go 1.15) task`. Maybe I can check if task name is capitalized then have a consistent task or Task. What do you think?

internal/executor/rpc/rpc.go
```diff
@@ -251,7 +268,7 @@ func (r *RPC) StreamLogs(stream api.CirrusCIService_StreamLogsServer) error {
 			currentTaskName = task.Name
 			currentCommand = x.Key.CommandName
 
-			streamLogger = r.logger.Scoped(task.UniqueName()).Scoped(currentCommand)
+			streamLogger := r.getCommandLogger(task, currentCommand)
```
edigaryev, line 271: If you look at the `ReportSingleCommand()`, it already retrieves and validates the existence of the command.

Perhaps we should do the same here, which would allow us greatly simplify the `getCommandLogger()` since it'll only accept a `build.Command`.

internal/executor/rpc/rpc.go
```diff
@@ -227,6 +226,24 @@ func (r *RPC) ReportSingleCommand(
 	return &api.ReportSingleCommandResponse{}, nil
 }
 
+func (r *RPC) getCommandLogger(task *build.Task, commandName string) *echelon.Logger {
+	commandLoggerScope := fmt.Sprintf("'%s'", commandName)
+	// make pretty names for some instruction types
+	command := task.GetCommand(commandName)
+	switch {
```
edigaryev, line 233: If we implement https://github.com/cirruslabs/cirrus-cli/pull/76/files#r489688469, we can simply use type assertion here.

internal/executor/build/task.go
```diff
@@ -89,11 +89,11 @@ func (task *Task) ProtoCommands() []*api.Command {
 	return result
 }
 
-func (task *Task) UniqueName() string {
```
edigaryev, line 92: Maybe we should simply keep this as is for now, and change later this once we figure out the `UniqueName()` implementation?

internal/executor/rpc/rpc.go
```diff
@@ -227,6 +226,24 @@ func (r *RPC) ReportSingleCommand(
...
+	case command == nil:
+		// no found :-(
+		break
+	case command.ProtoCommand.GetScriptInstruction() != nil:
+		commandLoggerScope += " script"
+	case command.ProtoCommand.GetBackgroundScriptInstruction() != nil:
+		commandLoggerScope += " background script"
+	case command.ProtoCommand.GetCacheInstruction() != nil:
+		commandLoggerScope += " cache"
+	}
```
edigaryev, line 243: I've ran this branch against CLI's `.cirrus.yml` and it seems that we can add the artifacts instruction here too.

internal/executor/rpc/rpc.go
fkorotkov: See 056773220f540420ab2b120da9101db961d1bae2

internal/executor/rpc/rpc.go
fkorotkov: see e250602468305ed08de4ed48ec72139611945162

internal/executor/rpc/rpc.go
fkorotkov: See 056773220f540420ab2b120da9101db961d1bae2

internal/executor/build/task.go
fkorotkov: I've added smart capitalization of `task`. PTAL 75ed0691276f2fa76f94fe6f31fc0d72c65b5a15

edigaryev (approved):

PR #77, opened 2020-09-10
Fall back to module version

fkorotkov:
If version is explicitly not configured


fkorotkov:
That won't work since `(devel)` is hard coded https://github.com/golang/go/issues/29228

edigaryev (approved):

PR #81, opened 2020-09-11
Introduce run's --docker-no-pull argument

edigaryev:
This brings back the old behavior where we always attempted to pull the latest container image before starting, but lets the user turn it off in case they want to run CI tasks with locally-built container images.

Always pulling images also avoids some tricky-to-diagnose cases like in cirruslabs/cirrus-ci-annotations#22, where running two similar configurations resulted in different results due to stale state in the Docker daemon.

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -209,7 +209,8 @@ func newRunCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "")
 
 	// Docker-related flags
-	cmd.PersistentFlags().BoolVar(&dockerNoPull, "docker-no-pull", false, "don't attempt to pull the images before starting containers")
+	cmd.PersistentFlags().BoolVar(&dockerNoPull, "docker-no-pull", false,
+		"don't attempt to pull the images before starting containers")
 
 	return cmd
 }
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/cirruslabs/cirrus-cli/internal/commands/logs"
 	"github.com/cirruslabs/cirrus-cli/internal/executor"
+	"github.com/cirruslabs/cirrus-cli/internal/executor/options"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/taskfilter"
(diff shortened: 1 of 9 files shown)


internal/commands/run.go
```diff
@@ -189,11 +201,15 @@ func newRunCmd() *cobra.Command {
...
 
+	// General flags
 	cmd.PersistentFlags().BoolVar(&dirty, "dirty", false, "if set the project directory will be mounted"+
 		"in read-write mode, otherwise the project directory files are copied, taking .gitignore into account")
 	cmd.PersistentFlags().StringArrayVarP(&environment, "environment", "e", []string{},
 		"set (-e A=B) or pass-through (-e A) an environment variable")
 	cmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "")
 
+	// Docker-related flags
+	cmd.PersistentFlags().BoolVar(&dockerNoPull, "docker-no-pull", false, "don't pull the images when starting containers")
```
fkorotkov, line 212:
```suggestion
	cmd.PersistentFlags().BoolVar(&dockerNoPull, "docker-no-pull", false, "don't attempt to pull the images before starting containers")
```

internal/executor/options/docker.go
```diff
@@ -0,0 +1,20 @@
+package options
+
+type DockerOptions struct {
+	NoPull       bool
+	NoPullImages []string
```
fkorotkov, line 5: `NoPullImages` sounds a bit weird IMO. What do you think about changing it yo `--docker-skip-pull`?

fkorotkov (approved):
LGTM. Left a small suggestion about the flag's description. Also left a comment of maybe changing `no` to `skip` but if you think `no` sounds goo than I don't mind merging as is.

internal/executor/options/docker.go
edigaryev: This is similar to the Ansible's flag: https://docs.ansible.com/ansible/latest/dev_guide/testing_integration.html#non-destructive-tests

I think it is a bit easier to type and I've also seen similarly composed clauses in other Docker commands:

```
% docker run --help | grep '\-\-no'
      --no-healthcheck                 Disable any container-specified HEALTHCHECK
```

```
% docker build --help | grep '\-\-no'
      --no-cache                Do not use cache when building the image
```

fkorotkov (approved):

PR #82, opened 2020-09-11
Support explicitly setting output format

### LINKED ISSUE #78: Flag to disable interactive output

fkorotkov:
Fixes #78

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -25,6 +25,7 @@ var ErrRun = errors.New("run failed")
 
 // General flags.
 var dirty bool
+var output string
 var environment []string
 var verbose bool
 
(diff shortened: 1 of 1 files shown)


internal/commands/run.go
```diff
@@ -194,6 +208,8 @@ func newRunCmd() *cobra.Command {
 	cmd.PersistentFlags().StringArrayVarP(&environment, "environment", "e", []string{},
 		"set (-e A=B) or pass-through (-e A) an environment variable")
 	cmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "")
+	cmd.PersistentFlags().StringVarP(&output, "output", "o", "auto", "output format of logs, "+
```
edigaryev, line 211: I think we need to show the default value here.

internal/commands/run.go
```diff
@@ -133,20 +134,33 @@ func run(cmd *cobra.Command, args []string) error {
 
 	var executorOpts []executor.Option
 
+	tryToDetectOutput := output == "auto"
+	if tryToDetectOutput && envVariableIsTrue("TRAVIS") {
+		output = "travis"
```
edigaryev, line 139: Perhaps we should make these strings constants, since they're used multiple times?

In this case they can also be re-used in the flag description.

internal/commands/run.go
fkorotkov: Should be fine, cobra shows it.

```bash
$: cirrus run --help
...
  -o, --output string             output format of logs, supported values: auto, interactive, simple, travis, github-actions (default "auto")
...
```

internal/commands/run.go
fkorotkov: Thought of it but Go doesn't have enums. :-(

edigaryev (approved):

PR #83, opened 2020-09-14
Flag to use local YAML parser instead of the remote one

### LINKED ISSUE #79: Flag to use local YAML parser instead of the remote one
Let's add `--experimental-local-parser` flag to use the experimental `parser` package instead of `rpcparser` that makes an RPC call to the Koltin backend.

edigaryev:
Resolves #79.


fkorotkov (approved):
LGMT once linting is passing

Issue #53, opened 2020-08-31
Productionize Startlark Configuration


fkorotkov:
Basic support for Startlark configuration was added in #48. We still need couple of things in order to productionize Starlark format to align with the [vision](https://twitter.com/fedor/status/1296462322530160641). Things like:

- [x] List of builtins (filesystem plus https://github.com/qri-io/starlib for now?)
- [x] Template loading
- [x] Template testing primitives
- [x] gRPC service for evaluating config
- [ ] Documentation

This will be an issue for tracking overall progress.

PR #85, opened 2020-09-15
Support loading Starlark modules from Git

edigaryev:
See #53.

Some things that still need to be taken care of:

* make Git-related storages bounded to a specific constant (e.g. 8 MiB)
https://github.com/cirruslabs/cirrus-cli/blob/04bbcae3addfcd5007460a26fad2e7353f40523d/pkg/larker/loader/git/git.go#L65-L66
* write a few more tests

### DIFF
--- a/pkg/larker/loader/git/git.go
+++ b/pkg/larker/loader/git/git.go
@@ -11,6 +11,7 @@ import (
 	"github.com/go-git/go-git/v5/plumbing/cache"
 	"github.com/go-git/go-git/v5/storage/filesystem"
 	"io/ioutil"
+	"os"
 	"regexp"
 )
 
@@ -20,7 +21,10 @@ type Locator struct {
 	Revision string
 }
 
-var ErrRetrievalFailed = errors.New("failed to retrieve a file from Git repository")
+var (
+	ErrRetrievalFailed = errors.New("failed to retrieve a file from Git repository")
+	ErrFileNotFound    = errors.New("file not found in a Git repository")
+)
 
 const (
 	// Captures the path after / in non-greedy manner.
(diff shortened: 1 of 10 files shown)


pkg/larker/loader/git/git.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := worktree.Checkout(&git.CheckoutOptions{
+		Hash: *hash,
+	}); err != nil {
+		return nil, err
+	}
+
+	// Read the file from the working tree
+	file, err := worktree.Filesystem.Open(locator.Path)
+	if err != nil {
+		return nil, err
```
fkorotkov, line 97: Will it have a pretty error message?

fkorotkov (commented):
I think it will be good to add a test or two about error messages when people will make typos like `.start` instead of `.star`. Otherwise LGTM so far.

edigaryev:
>make Git-related storages bounded to a specific constant (e.g. 8 MiB)

Implemented in https://github.com/cirruslabs/cirrus-cli/pull/85/commits/4e9fd367278d63c0d8685c2364bf2df9d1203f8b.

Unfortunately, the accounting is quite imprecise due to details hidden in `github.com/go-git/go-billy/v5/memfs` implementation. The most evident shortcomings of this are:

* we don't count intermediate files created as a result of calling `MkdirAll()` towards a limit
* we count all written bytes towards a limit and ignore the fact that `Write()`'s can overwrite the same file piece without increasing it's size

If at some point we want these to be fixed too, the next obvious step would be to submit the modifications to the https://github.com/go-git/go-billy itself.

fkorotkov:
Another option might be to store to a temp directory on disk without bounds and then delete it. Let's think about it more when implementing the microservice and caching of the 'load' results.

pkg/larker/loader/git/git.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/85/commits/cb79e3f39324792108e0741337ae800754cc4a39.

edigaryev:
> I think it will be good to add a test or two about error messages when people will make typos like `.start` instead of `.star`.

See https://github.com/cirruslabs/cirrus-cli/pull/85/commits/f8924a96f68d2dbeb96d396d130726345d1fdf87.

fkorotkov (approved):

PR #89, opened 2020-09-16
Basic GitHub filesystem support

edigaryev:
A gateway to proper filesystem builtins needed for #53.


pkg/larker/fs/github/github.go
```diff
@@ -0,0 +1,51 @@
...
+var ErrAPI = errors.New("failed to communicate with the GitHub API")
+
+type GitHub struct {
+	client    *github.Client
+	owner     string
+	repo      string
+	reference string
+}
+
+func New(owner, repo, reference, token string) *GitHub {
```
fkorotkov, line 21: `token` is not used.

pkg/larker/fs/github/github.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (gh *GitHub) Get(ctx context.Context, path string) ([]byte, error) {
+	fileContent, _, _, err := gh.client.Repositories.GetContents(ctx, gh.owner, gh.repo, path,
+		&github.RepositoryContentGetOptions{
+			Ref: gh.reference,
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrAPI, err)
```
fkorotkov, line 37: Maybe check if error is about a non-existing file and return `os.ErrNotExist`?

pkg/larker/fs/github/github.go
edigaryev: Indeed, fixed in https://github.com/cirruslabs/cirrus-cli/pull/89/commits/728ad2c32bd2817e3bc4a3a19cc0060849aec74a.

pkg/larker/fs/github/github.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/89/commits/aee354c515242d92e1d2e120b008bb7a57987f6a.

edigaryev:
> Maybe keep it `github.Client` to not create the client over and over again for every API call? `client` helper function can just return the right cleint in `New`.

Unfortunately `oauth2.NewClient()` requires a `context.Context` and it's not being transparent about whether it'll be used only for creation of the client or after some time later.

To avoid this ambiguity and to enable calling of `fs.FileSystem` methods with possibly different contexts I'm simply creating a new client for each method call.

Should be pretty cheap, huh?

fkorotkov (approved):

PR #96, opened 2020-09-20
cirrus module: exposed some Starlib builtins and added an env dict

edigaryev:
See #53.

### DIFF
--- a/pkg/larker/loader/loader.go
+++ b/pkg/larker/loader/loader.go
@@ -7,6 +7,10 @@ import (
 	"github.com/cirruslabs/cirrus-cli/pkg/larker/builtin"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker/fs"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker/loader/git"
+	"github.com/qri-io/starlib/encoding/json"
+	"github.com/qri-io/starlib/encoding/yaml"
+	"github.com/qri-io/starlib/hash"
+	"github.com/qri-io/starlib/http"
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarkstruct"
 	"os"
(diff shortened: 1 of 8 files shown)


pkg/larker/loader/loader.go
```diff
@@ -65,11 +71,28 @@ func (loader *Loader) LoadFunc() func(thread *starlark.Thread, module string) (s
...
 			return starlark.StringDict{
 				"fs": &starlarkstruct.Module{
 					Name:    "fs",
 					Members: builtin.FS(loader.ctx, loader.fs),
 				},
+				"env":  starlarkEnv,
+				"http": starlibModules["http"],
+				"hash": starlibModules["hash"],
+				"json": starlibModules["json"],
+				"yaml": starlibModules["yaml"],
```
fkorotkov, line 95: Let's also add `zipfile`, `re` and `base64. Can do here or in a separate PR.

pkg/larker/loader/loader.go
```diff
@@ -65,11 +71,28 @@ func (loader *Loader) LoadFunc() func(thread *starlark.Thread, module string) (s
...
+					return nil, err
+				}
+			}
+
 			return starlark.StringDict{
 				"fs": &starlarkstruct.Module{
 					Name:    "fs",
 					Members: builtin.FS(loader.ctx, loader.fs),
 				},
+				"env":  starlarkEnv,
```
fkorotkov, line 91: Seems we can rename `starlibModules` to `builtinModules` and then do something like:

```python
builtinModules["fs"] = ...
builtinModules["env"] = ...
return builtinModules
```

This way there is no need for duplicated `"http"`, `"hash"`, etc. What do you think?

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/96/commits/9645b0411b06fbfe0062c77cc16d3f023fae8734.

pkg/larker/loader/loader.go
edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/96/commits/40d9ac23705bfac15525e21c0c14fae01b6909f8.

fkorotkov (approved):

PR #98, opened 2020-09-21
Better error message on Docker bridge failure

fkorotkov:


internal/executor/rpc/rpc.go
```diff
@@ -140,6 +140,9 @@ func (r *RPC) Start(ctx context.Context) error {
 
 	listener, err := net.Listen("tcp", address)
 	if err != nil {
+		if strings.Contains(err.Error(), "bind: cannot assign requested address") {
```
edigaryev, line 143: I don't think this will work for non-english locales. A better way would be to check for `errors.Is(err, syscall.EADDRNOTAVAIL)`.

However, both won't probably work on Windows anyway.

internal/executor/rpc/rpc.go
```diff
@@ -140,6 +140,9 @@ func (r *RPC) Start(ctx context.Context) error {
...
+			return fmt.Errorf("%w: failed to connect to docker network bridge (is docker running?)", ErrRPCFailed)
```
edigaryev, line 144:
```suggestion
			return fmt.Errorf("%w: failed to assign Docker network bridge address %s (is Docker running?)", ErrRPCFailed, address)
```

internal/executor/rpc/rpc.go
fkorotkov: Seems so but at least on majority platforms the error message will be better.

edigaryev (approved):

PR #101, opened 2020-09-22
Task filtering by labels

### LINKED ISSUE #63: cirrus run should allow to filter by labels
Related to #62. When there are bunch of tasks with the same name but different labels (container image, some env variables) it not possible to target a particular task. `cirrus run` should be able to match by `UniqueName` too.

edigaryev:
Resolves #63.

### DIFF
--- a/internal/executor/taskfilter/taskfilter.go
+++ b/internal/executor/taskfilter/taskfilter.go
@@ -63,8 +63,10 @@ func extractLabels(s string) (result []string) {
 	return
 }
 
-func matchLabels(soughtLabels, actualLabels []string) (numMatchedLabels int) {
-	for _, soughtLabel := range soughtLabels {
+func containsAll(desiredLabels, actualLabels []string) bool {
+	var numMatchedLabels int
+
+	for _, soughtLabel := range desiredLabels {
 		for _, actualLabel := range actualLabels {
 			if strings.EqualFold(soughtLabel, actualLabel) {
 				numMatchedLabels++
@@ -73,9 +75,5 @@ func matchLabels(soughtLabels, actualLabels []string) (numMatchedLabels int) {
 		}
 	}
 
-	return
-}
-
-func containsAll(desiredLabels, actualLabels []string) bool {
-	return matchLabels(desiredLabels, actualLabels) == len(desiredLabels)
+	return numMatchedLabels == len(desiredLabels)
 }
--- a/internal/executor/taskfilter/taskfilter.go
+++ b/internal/executor/taskfilter/taskfilter.go
@@ -34,7 +34,7 @@ func MatchExactTask(desiredTaskName string) TaskFilter {
 				actualLabels = task.Metadata.UniqueLabels
 			}
 
-			if !containsAll(desiredLabels, actualLabels) {
+			if !containsAll(actualLabels, desiredLabels) {
 				continue
 			}
 
@@ -63,7 +63,7 @@ func extractLabels(s string) (result []string) {
 	return
 }
 
-func containsAll(desiredLabels, actualLabels []string) bool {
+func containsAll(actualLabels []string, desiredLabels []string) bool {
 	var numMatchedLabels int
 
 	for _, desiredLabel := range desiredLabels {
--- a/internal/executor/taskfilter/taskfilter.go
+++ b/internal/executor/taskfilter/taskfilter.go
@@ -66,9 +66,9 @@ func extractLabels(s string) (result []string) {
 func containsAll(desiredLabels, actualLabels []string) bool {
 	var numMatchedLabels int
 
-	for _, soughtLabel := range desiredLabels {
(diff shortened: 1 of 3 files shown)


internal/executor/taskfilter/taskfilter.go
```diff
@@ -18,14 +18,61 @@ func MatchExactTask(desiredTaskName string) TaskFilter {
...
+			// In case we're looking for a task with specific labels — extract them and ensure they all match
+			desiredLabels := extractLabels(strings.TrimPrefix(desiredTaskNameLower, taskNameLower))
+
+			var numMatchedLabels int
+
+			if task.Metadata != nil {
+				numMatchedLabels = matchLabels(desiredLabels, task.Metadata.UniqueLabels)
+			}
+
+			if numMatchedLabels != len(desiredLabels) {
```
fkorotkov, line 38: Maybe create a util method `containsAll` and have a single condition `containsAll(task.Metadata.UniqueLabels, desiredLabels)`?

edigaryev: Wouldn't you still have two conditions, since we need to check for `task.Metadata != nil`?

fkorotkov: that's true but still IMO the code will be more readable than comparing how many labels matched. 🤔

edigaryev: See b11cfb1e092c8f1b2407d24c01b581a13e04a2ea.

internal/executor/taskfilter/taskfilter.go
```diff
@@ -18,14 +18,64 @@ func MatchExactTask(desiredTaskName string) TaskFilter {
...
+			}
+		}
+	}
+
+	return
+}
+
+func containsAll(desiredLabels, actualLabels []string) bool {
+	return matchLabels(desiredLabels, actualLabels) == len(desiredLabels)
+}
```
fkorotkov, line 81: I meant something like:

```suggestion
func containsAll(values []string, items []string) bool {
	ItemsLoop:
	for _, item := range items {
		for _, value := range values {
			if strings.EqualFold(value, item) {
				continue ItemsLoop
			}
		}
		return false
	}
	return true
}
```

internal/executor/taskfilter/taskfilter.go
```diff
@@ -18,14 +18,64 @@ func MatchExactTask(desiredTaskName string) TaskFilter {
...
+
+			// In case we're looking for a task with specific labels — extract them and ensure they all match
+			desiredLabels := extractLabels(strings.TrimPrefix(desiredTaskNameLower, taskNameLower))
+
+			var actualLabels []string
+			if task.Metadata != nil {
+				actualLabels = task.Metadata.UniqueLabels
+			}
+
+			if !containsAll(desiredLabels, actualLabels) {
```
fkorotkov, line 37: This should be also fixed with the below suggestion

```suggestion
			if !containsAll(actualLabels, desiredLabels) {
```

internal/executor/taskfilter/taskfilter.go
edigaryev: I don't really like the continue label since it needlessly breaks the program's structure.

However I've moved everything into a single function in b6ce7b0.

Does that work for you?

internal/executor/taskfilter/taskfilter.go
```diff
@@ -63,8 +63,10 @@ func extractLabels(s string) (result []string) {
 	return
 }
 
-func matchLabels(soughtLabels, actualLabels []string) (numMatchedLabels int) {
-	for _, soughtLabel := range soughtLabels {
+func containsAll(desiredLabels, actualLabels []string) bool {
+	var numMatchedLabels int
+
+	for _, soughtLabel := range desiredLabels {
```
fkorotkov, line 69: Maybe rename to `desiredLabel` for consistency?

internal/executor/taskfilter/taskfilter.go
```diff
@@ -63,8 +63,10 @@ func extractLabels(s string) (result []string) {
 	return
 }
 
-func matchLabels(soughtLabels, actualLabels []string) (numMatchedLabels int) {
-	for _, soughtLabel := range soughtLabels {
+func containsAll(desiredLabels, actualLabels []string) bool {
```
fkorotkov, line 66: How about still switching places of the arguments so the slice where the search will be performed will be first.

internal/executor/taskfilter/taskfilter.go
edigaryev: See 68552bc.

internal/executor/taskfilter/taskfilter.go
edigaryev: See ac0909fc56b31e7ca3016e3aa9f0600eb120521e.

fkorotkov (approved):

PR #103, opened 2020-09-22
[skip-ci] Document all run options

fkorotkov:


README.md
```diff
@@ -55,12 +55,23 @@ To run Cirrus tasks, simply switch to a directory where the `.cirrus.yml` is loc
...
+It is also possible to run a particular task by name:
                           
 ```shell script
 cirrus run "Tests (Go 1.15)"
 ```
 
+Or pass some extra environment variables with `-e` flag:
+
+```shell script
+cirrus run -e TEST_PATTER="*Integration*" "Tests (Go 1.15)"
```
edigaryev, line 72: >`TEST_PATTER`

🤔

It'd probably be better if this will be some real-word example.

README.md
```diff
@@ -55,12 +55,23 @@ To run Cirrus tasks, simply switch to a directory where the `.cirrus.yml` is loc
 cirrus run
 ```
 
-It is also possible to run a task by name:
+By default, working directory will be `rsync`ed into a container while respecting `.gitignore`
```
edigaryev, line 58: I think we should move the `--dirty` part of this paragraph after the `cirrus run` example below and provide it's own example to simplify reading.

README.md
```diff
@@ -55,12 +55,23 @@ To run Cirrus tasks, simply switch to a directory where the `.cirrus.yml` is loc
...
+configuration. This makes sure Cirrus Tasks are execution from a clean state only with source code
```
edigaryev, line 59:
```suggestion
configuration. This makes sure Cirrus Tasks are executed from a clean state only with source code
```

README.md
fkorotkov: A suggestion will be appreciated

fkorotkov: Add an example on testing a release by overriding `CIRRUS_TAG`

README.md
```diff
@@ -55,12 +55,25 @@ To run Cirrus tasks, simply switch to a directory where the `.cirrus.yml` is loc
 cirrus run
 ```
 
-It is also possible to run a task by name:
+By default, working directory will be `rsync`ed into a container while respecting `.gitignore`
+configuration. This makes sure Cirrus Tasks are executed from a clean state only with source code
+changes.
+
+There is an option to directly mount working directory as is with `--dirty` flag which
```
edigaryev, line 62: We should probably explain "read-only" in this paragraph better, since it makes little sense with `--dirty`, which is actually a mode in which writes are done against the original project directory.

How about replacing this paragraph with:

>In case `rsync`-ing the whole working directory is too costly, you can pass a `--dirty` flag which will result in all operations being done against the actual working directory (and not it's copy):

>```shell script
>cirrus run --dirty Lint
>```

>Since most linters and code-analysis tools are read-only by their nature — greater speed is achieved at the cost of additional isolation.

fkorotkov: How about:

>  (and not it's `rsync`ed copy)

> Since most linters and code-analysis tools are read-only by their nature there is no need in extra precautions and the potentially costly `rsync`-ing can be safely avoided.

edigaryev: OK.

fkorotkov: See 5b6d7bc

edigaryev (approved):

Issue #59, opened 2020-09-03
Complete the parser


edigaryev:
#58 introduces a YAML parser that's able to run most of the CLI tasks without network connection.

However, this results in at least two parsers in-the-wild (the other being the one that processes `.cirrus.yml` in the cloud). To avoid the divergence in the future, it might be a good idea to consolidate everything as a CLI package, which is available for use in any Go program.

To do this, the following things need to be implemented:

* support all execution environments
* support JSON Schema
* integrate already existing tests

fkorotkov:
Almost everything done except JSON schema. Created a separate issue https://github.com/cirruslabs/cirrus-cli/issues/148

PR #108, opened 2020-09-23
Introduce TestViaRPC to converge parser and rpcparser's implementations

edigaryev:
A single test for now.

Had to switch the `parser` from `gopkg.in/yaml.v2` to `github.com/goccy/go-yaml` since the former doesn't support aliases:

* https://github.com/go-yaml/yaml/issues/391
* https://github.com/go-yaml/yaml/issues/624

See #59.


fkorotkov (commented):
Looks awesome!

We are going to eventually remove the RPC parser so let's just commit the expected parsed results as a JSON dump next to the YAML files. And in a test we can have a check that if `.json` file doesn't exist we call the RPC and create the expected result. This way it will be easy to generate the test data and eventually just remove the RPC parser.

edigaryev:
> We are going to eventually remove the RPC parser so let's just commit the expected parsed results as a JSON dump next to the YAML files. And in a test we can have a check that if `.json` file doesn't exist we call the RPC and create the expected result. This way it will be easy to generate the test data and eventually just remove the RPC parser.

Implemented in https://github.com/cirruslabs/cirrus-cli/pull/108/commits/28762f39fb21625f74f1c5c9aed3080c4e823efc.

fkorotkov (approved):

PR #111, opened 2020-09-26
Generic Proto Instance Parser

fkorotkov:
This an implementation of a parser that can take a type descriptor (basically a serializable definition) and parse it back into a proto message. This is need for the future of the CLI when it will be able to act a service for evaluating of the YAML and Starlark configs for the Cirrus Cloud. `EvaluateConfigRequest` will need to be modified: `AdditionalInstancesInfo` will be added.

```protobuf
message EvaluateConfigRequest {
  message AdditionalInstancesInfo {
    map<string, string> instances = 1; // field name to proto FQN
    google.protobuf.FileDescriptorSet descriptor = 2;
  }

  string yaml_config = 1;
  string starlark_config = 2;
  map<string, string> environment = 3;
  repeated string affected_files = 4;
  map<string, string> files_contents = 5;
  AdditionalInstancesInfo additional_instances_info = 6;
}
```

### DIFF
--- a/pkg/parser/instance/proto.go
+++ b/pkg/parser/instance/proto.go
@@ -1,0 +1,206 @@
+package instance
+
+import (
+	"github.com/cirruslabs/cirrus-cli/pkg/parser/nameable"
+	"github.com/cirruslabs/cirrus-cli/pkg/parser/node"
+	"github.com/cirruslabs/cirrus-cli/pkg/parser/parseable"
+	"github.com/cirruslabs/cirrus-cli/pkg/parser/schema"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/dynamicpb"
+	"strconv"
+	"strings"
+)
+
+type ProtoInstance struct {
+	proto *dynamicpb.Message
+
+	parseable.DefaultParser
+}
+
+//nolint:gocognit // it's a parser, there is a lot of boilerplate
+func NewProtoParser(desc protoreflect.MessageDescriptor, mergedEnv map[string]string) *ProtoInstance {
+	instance := &ProtoInstance{
+		proto: dynamicpb.NewMessage(desc),
+	}
+
+	fields := desc.Fields()
+	for i := 0; i < fields.Len(); i++ {
+		field := fields.Get(i)
+		fieldName := string(field.Name())
+		switch field.Kind() {
+		case protoreflect.MessageKind:
+			instance.OptionalField(nameable.NewSimpleNameable(fieldName), schema.TodoSchema, func(node *node.Node) error {
+				if field.IsMap() {
+					fieldInstance := instance.proto.NewField(field)
+					mapping, err := node.GetStringMapping()
+					if err != nil {
+						return err
+					}
+					for key, value := range mapping {
+						fieldInstance.Map().Set(
+							protoreflect.ValueOfString(key).MapKey(),
+							protoreflect.ValueOfString(value),
+						)
+					}
+					instance.proto.Set(field, fieldInstance)
+					return nil
+				} else if field.IsList() {
+					fieldInstance := instance.proto.NewField(field)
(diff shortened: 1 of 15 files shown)


pkg/parser/instance/proto.go
```diff
@@ -0,0 +1,141 @@
...
+				return nil
+			})
+		case protoreflect.Int64Kind, protoreflect.Sint64Kind, protoreflect.Uint64Kind,
+			protoreflect.Fixed64Kind, protoreflect.Sfixed64Kind:
+			instance.OptionalField(nameable.NewSimpleNameable(field.JSONName()), schema.TodoSchema, func(node *node.Node) error {
+				value, err := node.GetExpandedStringValue(mergedEnv)
+				if err != nil {
+					return err
+				}
+				parsedValue, err := strconv.ParseInt(value, 10, 64)
```
edigaryev, line 66: Shouldn't that be in `else` branch of the following `if`?

pkg/parser/testdata/proto-instance.yml
```diff
@@ -0,0 +1,22 @@
...
+      env:
+        MYSQL_ROOT_PASSWORD: ""
+
+regular_task:
+  container:
+    <<: *container_body
+
+proto_task:
+  container:
+    <<: *container_body
```
edigaryev, line 22: Looks like the newline is missing here.

pkg/parser/instance/proto.go
```diff
@@ -0,0 +1,141 @@
...
+				}
+				parsedValue, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					return err
+				}
+				instance.proto.Set(field, protoreflect.ValueOfFloat64(parsedValue))
+				return nil
+			})
+		case protoreflect.GroupKind, protoreflect.BytesKind:
+			// not supported
```
edigaryev, line 129: Perhaps we should just `case default` here instead of specific types for clarity?

fkorotkov: IMO in this case it's better to explicitly declare two unsupported cases. If in future there will some new kinds then linter will make sure they'll be covered. 🤷‍♂️

pkg/parser/testdata/proto-instance.yml
fkorotkov: See c4ef091

pkg/parser/instance/proto.go
fkorotkov: See c47a3ef

pkg/parser/testdata/proto-instance.yml
```diff
@@ -0,0 +1,22 @@
...
+      cpu: 1
+      memory: 1024
+      env:
+        MYSQL_ROOT_PASSWORD: ""
+
+regular_task:
+  container:
+    <<: *container_body
+
+proto_task:
```
edigaryev, line 20: What's the difference between two of these tasks?

fkorotkov: One is parsed via the regular parser and the `proto` one with the new generic one. And they yield the same results!!!

edigaryev: Besides the fact that `proto-instance.json` is parsed with the RPC parser and `proto-instance.yml` with the Go parser — these two tasks (`regular` and `proto`) look essentially the same and seem to be treated equally, why duplicate them?

fkorotkov: This is the Go YAML `parser` package. I can leave only `proto` task but in case there will be some changes in the regular YAML parser of `container` instances it will be great to see this test to fail and make sure that `proto_container` will keep in sync.

fkorotkov:
In order to fix the parser I had to add two ifs while handling `MessageKind`: one for `repeated` lists and one for `map`s.

edigaryev (approved):

PR #113, opened 2020-09-29
More RPC-parser-powered tests: matrix and pipe

edigaryev:
More to come, merge at will.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strconv"
 )
 
 var (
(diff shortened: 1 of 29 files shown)


pkg/parser/parser.go
```diff
@@ -333,33 +346,60 @@ func (p *Parser) createServiceTasks(protoTasks []*api.Task) ([]*api.Task, error)
...
 						ScriptInstruction: &api.ScriptInstruction{
-							Scripts: []string{"true"},
+							Scripts: []string{fmt.Sprintf("gcloud docker -- push gcr.io/cirrus-ci-community/%s:latest",
+								dockerfileHash)},
 						},
 					},
 				},
 			},
+			Environment: map[string]string{
+				"DISPLAY": ":99.0",
```
fkorotkov, line 384: I don't think it's there by default.

pkg/parser/parser.go
```diff
@@ -441,8 +481,53 @@ func populateUniqueLabels(task *api.Task) error {
...
+	// Sort labels to make them comparable
+	sort.Strings(labels)
+
 	// Update task
 	task.Metadata.UniqueLabels = labels
 
 	return nil
 }
+
+func removeCommonLabels(tasks []*api.Task) {
```
fkorotkov, line 498: Can `populateUniqueLabels` create the correct set of labels in the first place rather then here trying to clean it up?

edigaryev: I thought about that too, but the problem these functions are trying to solve is two-pass its nature because you can't know which labels will be assigned in advance, so you assign them first and then optimize.

Do you think moving `populateUniqueLabels()` functionality into `removeCommonLabels()` and renaming it to reflect that afterwards would be a better idea?

pkg/parser/parser.go
edigaryev: Indeed 🤦‍♂️

Fixed in https://github.com/cirruslabs/cirrus-cli/pull/113/commits/cc8755b55854fb4a51c1c5e37ba4ed04a26aef5d.

pkg/parser/parser.go
fkorotkov: I think `populateUniqueLabels` should accept all tasks and not just one and populate unique labels for all of them similar to how `TaskLabelUtils#uniqueLabels` works in the cloud.

edigaryev: I see, does https://github.com/cirruslabs/cirrus-cli/pull/113/commits/fc005e3385e8c469f8effc8b5e954ff60aac88ea work for you?

fkorotkov: Yes 😍

fkorotkov (approved):

PR #114, opened 2020-10-01
Introduce serve command

edigaryev:
See #53.

### DIFF
--- a/internal/evaluator/evaluator.go
+++ b/internal/evaluator/evaluator.go
@@ -75,10 +75,10 @@ func evaluateConfig(ctx context.Context, request *api.EvaluateConfigRequest) (*a
 		yamlConfigs = append(yamlConfigs, request.YamlConfig)
 	}
 
+	fs := fsFromEnvironment(request.Environment)
+
 	// Run Starlark script and register generated YAML configuration (if any)
 	if request.StarlarkConfig != "" {
-		fs := fsFromEnvironment(request.Environment)
-
 		lrk := larker.New(
 			larker.WithFileSystem(fs),
 			larker.WithEnvironment(request.Environment),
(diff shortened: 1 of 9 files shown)


internal/evaluator/evaluator.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		yamlConfigs = append(yamlConfigs, generatedYamlConfig)
+	}
+
+	// Parse combined YAML
+	p := parser.New(
+		parser.WithEnvironment(request.Environment),
+		parser.WithFilesContents(request.FilesContents),
```
fkorotkov, line 65: Let's also add GitHub FS in case environment contains `CIRRUS_REPO_CLONE_TOKEN`, `CIRRUS_REPO_NAME`, `CIRRUS_REPO_OWNER` and `CIRRUS_CHANGE_IN_REPO`. Seems this will require the parser to check explicit file contents first and if a file is not presented then check in fs.

edigaryev: Implemented in https://github.com/cirruslabs/cirrus-cli/pull/114/commits/258567df788f30780ecc1894c8c3cc8f96f400c7.

I'm not sure why we need to touch files contents or parser at all, though, since GitHub FS is only used by Starlark.

fkorotkov (commented):
I think `ContentHash` should also check with `fs` if `filesContents` doesn't contain the required path.

edigaryev:
> I think `ContentHash` should also check with `fs` if `filesContents` doesn't contain the required path.

PTAL at https://github.com/cirruslabs/cirrus-cli/pull/114/commits/3cdbd63ce9579a3fb83dfeef368198c5cd3abded.

fkorotkov (approved):

PR #115, opened 2020-10-02
Introduce internal test command

### LINKED ISSUE #95: Cirrus Template Testing
Since now you can create configurations in Starlark we need to think about how to test them.

A simple template repository looks like:

```
lib.star
```

One thing we can do it to have `tests` folder with test projects and check that evaluation the template yielded expected tasks.

```
tests/foo-project/.cirrus.env # optional file with environment variables for evaluation
tests/foo-project/.cirrus.star # will contain load("../../lib.star", "foo") and main for testing
tests/foo-project/.cirrus.expected.yml # expected result to diff agains
tests/foo-project/package.json # some other files for testing `fs` for example
lib.star
```

Then we can have a separate internal subcommand like `cirrus internal test tests/*` that will take a list of folders and run evaluation and checking against the expected YML.

edigaryev:
Resolves #95.

### DIFF
--- a/internal/commands/internal/test.go
+++ b/internal/commands/internal/test.go
@@ -6,7 +6,6 @@ import (
 	"github.com/cirruslabs/cirrus-cli/internal/commands/logs"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker/fs/local"
-	"github.com/cirruslabs/echelon/renderers"
 	"github.com/go-test/deep"
 	"github.com/goccy/go-yaml"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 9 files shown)


internal/commands/internal/test.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Configure hierarchical progress renderer
+	renderer := renderers.NewInteractiveRenderer(os.Stdout, nil)
```
fkorotkov, line 42: I don't think you need this here since you are using `logs.GetLogger` in the next lines.

fkorotkov (approved):
LGTM. Seems find an unused code thought.

internal/commands/internal/test.go
edigaryev: Good catch, removed in https://github.com/cirruslabs/cirrus-cli/pull/115/commits/670898f66e88fa8290b606b097ccd624b7027ca9.

Issue #57, opened 2020-09-02
Add instructions how to run on Google Cloud Build


fkorotkov:
Ideally includuing using an http cache to store on Google Cloud Storage.

PR #120, opened 2020-10-06
Fallback to connector mode in case we fail to bind to Docker's bridge

edigaryev:
This makes running on Google Cloud Build (#57) and similar configurations (when CLI is started inside of a container, but Docker is nevertheless available) possible and potentially fixes a lot of other corner cases when we fail to detect Docker's bridge IP properly.

Note that this requires two more things before this gets merged:

* proper go.mod reference to github.com/cirruslabs/cirrus-ci-agent
* proper AgentImage reference in volume.go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,12 +10,12 @@ require (
 	github.com/Microsoft/go-winio v0.4.14 // indirect
 	github.com/PaesslerAG/gval v1.1.0
 	github.com/certifi/gocertifi v0.0.0-20200922220541-2c3bb06c6054
-	github.com/cirruslabs/cirrus-ci-agent v1.11.0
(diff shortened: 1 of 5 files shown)


edigaryev:
I've actually tried running these changes on Cloud Build and realized that even through we can now connect to the agent  from the CLI (and not the other way around as it is right now), this still wouldn't work since CLI is already running inside of a container and it's `127.0.0.1` belongs to a separate namespace than the host's `127.0.0.1` on which we expose the agent's connector port.

And even through this still can be worked around by creating a Unix socket, mounting it into the container and communicating over it instead of TCP, let's see if #121 works instead, since it's actually even simpler assuming we only want to support Cloud Build for now.

PR #129, opened 2020-10-12
Pick a port for serving from environment

fkorotkov:
Plus allow to pass a port via an environment variable


edigaryev:
I don't quite understand why we need such defaults.

`cirrus serve` is for the internal use primarily, and as a user running CLI Docker container I'd rather get a `--help` output than any not-so-helpful message from `cirrus serve`.

fkorotkov:
@edigaryev agreed, let's keep the port part since it is nicer for running the service in Cloud Run.

edigaryev (approved):

Issue #124, opened 2020-10-10
task script "fails" even though it had exit code 0?


duckinator:
When running cirrus run --verbose --dirty lint[1] in a fresh clone of https://github.com/awooos/flail on my system, it says `container exited with <nil> error and exit code 0`, then tells me it failed.

I suspect having logs from the actual running commands would help, but I'm not sure if/how I can get those.

```
~/dev/os/awooos/flail$ cirrus run --verbose --dirty lint
🕙 'Lint' Task 02:45
   ✅ Preparing execution environment... 3.8s
   ✅ docker pull 6.5s
   ✅ 'install' script 02:27
   ❌ 'main' script 1.0s
      begin streaming logs
      command failed

   creating Docker client
   creating container using working volume cirrus-working-volume-9ca71a13-5127-4bd1-bc50-ed060f8b3519
   starting container 0b05f1ae00845dc2b8012cff07494d31a01f4929380f6d98f87bec0b7b2e8635
   received heartbeat
   agent signal: urgent I/O condition
   received heartbeat
   waiting for container 0b05f1ae00845dc2b8012cff07494d31a01f4929380f6d98f87bec0b7b2e8635 to finish
   received heartbeat
   received heartbeat
   container exited with <nil> error and exit code 0
   cleaning up container 0b05f1ae00845dc2b8012cff07494d31a01f4929380f6d98f87bec0b7b2e8635

Error: build failed: task Lint (0) failed
2020/10/10 19:38:27 build failed: task Lint (0) failed
~/dev/os/awooos/flail$
```

duckinator:
I just noticed that it didn't clean up one of the containers it used, and looked into that a bit.

It appears to be a file permission problem. I do know that on Fedora, you frequently have to use the `:z` for volume bind mounts (see https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label ). That might be related?

```
$ docker ps -a
CONTAINER ID        IMAGE                                               COMMAND                  CREATED             STATUS                      PORTS               NAMES
816d478bcad2        gcr.io/cirrus-ci-community/cirrus-ci-agent:v1.8.0   "/bin/sh -c 'cp /bin…"   5 minutes ago       Exited (23) 5 minutes ago                       reverent_mcclintock
$ docker logs 816d478bcad2
rsync: opendir "/project-dir/." failed: Permission denied (13)
[...]

PR #131, opened 2020-10-13
Disable SELinux for service container and --dirty containers

edigaryev:
Unfortunately, taking the least-privilege path by running the container with the current users context (using `--security-opt="label=user:USER"`) and other labels doesn't work, because the container filesystem seems to be already labeled with the container-specific labels, which prevents us from even running the entrypoint binary.

See #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,8 @@ cirrus run --environment CIRRUS_HTTP_CACHE_HOST=http-cache-host.internal:8080
 Cirrus CLI tries to run in different environments, but in some environments we choose to provide more usability at the cost of a few security trade-offs:
 
 * SELinux
-  * service container that copies the project directory into a per-task Docker volume using `rsync` always runs unconfined
   * task container runs unconfined only if the `--dirty` flag is used
+  * service container that copies the project directory into a per-task Docker volume using `rsync` always runs unconfined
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -111,6 +111,16 @@ To start using your own HTTP caching server simply pass it's hostname as `CIRRUS
 cirrus run --environment CIRRUS_HTTP_CACHE_HOST=http-cache-host.internal:8080
 ```
 
+## Security
+
+Cirrus CLI tries to run in different environments, but in some environments we choose to provide more usability at the cost of a few security trade-offs:
+
+* SELinux
+  * service container that copies the project directory into a per-task Docker volume using `rsync` always runs unconfined
```
fkorotkov, line 119: Do we even need to mention the `rsync` internal and this path? The volume creation part is fully controlled by the CLI and IMO a user intents to provide contents of the working directory to the containers. Maybe only mention caveats for the dirty mode where user defined code will be run agains a mounted directory?

edigaryev: I think this is good for transparency, since people who will be reading this SELinux section might be concerned about what actually runs unconfined.

And mentioning `rsync` here helps to explain a possible attack surface.

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/131/commits/02221a1880f1aebbda7077a0ebf437b16524f182.

fkorotkov (approved):

PR #134, opened 2020-10-14
Support optional name field in the artifacts and cache commands

edigaryev:


fkorotkov:
Was there a test that was failing? We should probably add the YAML test file then. 🤔

edigaryev:
> Was there a test that was failing? We should probably add the YAML test file then. 🤔

No, it's just a missing piece I've noted while adding RPC tests.

fkorotkov (approved):

PR #136, opened 2020-10-16
Respect additional instances while serving

fkorotkov:
Make sure additional instances are respected and since proto messages are not getting added to the global proto type registry handle `NotFound` errors properly.

Still TBD how to generate proper labels for additional instances and how to generate docker builders for GKEContianer instances with `dockerfile` set. But theses two issues are for a separate PR.

### DIFF
--- a/internal/evaluator/evaluator.go
+++ b/internal/evaluator/evaluator.go
@@ -2,6 +2,7 @@ package evaluator
 
 import (
 	"context"
+	"fmt"
 	"github.com/cirruslabs/cirrus-ci-agent/api"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker/fs"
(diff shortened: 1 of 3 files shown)


fkorotkov:
@edigaryev fixed the NPE and it's ready for a review. The GKE issue I'll resolve on the cloud side by creating Docker Builder tasks there. Also will unify labels between CLI parser and the Cloud but in a separate PR.

internal/evaluator/evaluator.go
```diff
@@ -98,10 +101,30 @@ func (r *ConfigurationEvaluatorServiceServer) EvaluateConfig(
 		yamlConfigs = append(yamlConfigs, generatedYamlConfig)
 	}
 
+	additionalInstances := make(map[string]protoreflect.MessageDescriptor)
```
edigaryev, line 104: It would be nice if this added piece of code would provide an explanation of what it's doing and some test, since it's somewhat complex and isn't covered at all.

fkorotkov: Added a test on the CLI side in 9795392c9c1b4abcdd2d8a4d3e8d66dd206a70a4

edigaryev (approved):

PR #138, opened 2020-10-29
Parsing improvements

fkorotkov:
This is a set of changes I had to do in order to make sure all parsing tests in Cirrus Cloud are passing agains the CLI in `serve` mode. I tried to break them in more or less independent commits but there were a few where I updated test data (added new fields) which can be safely ignored.


pkg/parser/parser.go
```diff
@@ -374,7 +374,10 @@ func (p *Parser) createServiceTask(
 		Metadata: &api.Task_Metadata{
 			Properties: environment.Merge(
 				task.DefaultTaskProperties(),
-				map[string]string{"skip_notifications": "true"},
+				map[string]string{
+					"skip_notifications": "true",
+					"auto_cancellation":  protoTask.Metadata.Properties["auto_cancellation"],
```
edigaryev, line 380: What if the `auto_cancellation` key is not set in `protoTask.Metadata.Properties`?

pkg/parser/parser_test.go
```diff
@@ -209,9 +209,7 @@ func TestViaRPCInvalid(t *testing.T) {
 		Message string
 	}{
 		{"validation-badDependencies.yml", "error in dependencies between tasks: b, c, d"},
-		{"validation-badDependencyNames.yml", "there's no task 'fooo', but task 'bar' depends on it"},
```
edigaryev, line 212: Why remove these?

pkg/parser/task/task.go
```diff
@@ -56,34 +60,38 @@ func NewTask(
...
-			if err != nil {
-				return err
-			}
-			task.proto.Instance = anyInstance
+				anyInstance, err := ptypes.MarshalAny(containerInstance)
+				if err != nil {
+					return err
+				}
+				task.proto.Instance = anyInstance
+				task.proto.Environment = environment.Merge(task.proto.Environment, map[string]string{"CIRRUS_OS": "linux"})
```
edigaryev, line 81: How does this relate to the `env["CIRRUS_OS"]` above and why is it always `linux`?

fkorotkov: Right, can populate though `platform` field. See 445dbb7ba7f982fc57628affc02a988cdfa4eba4

edigaryev (approved):

PR #139, opened 2020-10-30
Persistent worker mode

edigaryev:
The simplest possible invocation with semi-automatic worker name generation:

```
cirrus worker run --token <pool registration token obtained from the web UI>
```


internal/executor/cache/cache.go
```diff
@@ -32,7 +32,7 @@ func New(dir string, namespace string) (*Cache, error) {
 		dir = userCacheDir
 	}
 
-	namespaceDir := filepath.Join(dir, "cirrus", namespace)
+	namespaceDir := filepath.Join(dir, "cirrus", "projects", namespace)
```
fkorotkov, line 35: This will invalidate the existing caches. But should be OK.

internal/executor/instance/volume.go
```diff
@@ -20,9 +20,12 @@ var (
 	ErrVolumeCleanupFailed  = errors.New("failed to clean up working volume")
 )
 
+// AgentImage is the image we'll use to create a working volume.
+var AgentImage = "gcr.io/cirrus-ci-community/cirrus-ci-agent:v" + AgentVersion
```
fkorotkov, line 24: why `var`? can probably out in const after `AgentVersion`.

edigaryev: Indeed, fixed in https://github.com/cirruslabs/cirrus-cli/pull/139/commits/5ddd4af7dd388e6098b0e3e541016a1d8e0f81b4.

internal/executor/instance/persistentworker_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// Does it cache the agent?
+	cachedRetrievalStart := time.Now()
+
+	_, err = instance.RetrieveAgentBinary(context.Background(), instance.AgentVersion, runtime.GOOS, runtime.GOARCH)
+	if err != nil {
+		t.Fatal()
+	}
+
+	assert.WithinDuration(t, cachedRetrievalStart, time.Now(), 100*time.Millisecond)
```
fkorotkov, line 27: Maybe check that file creation timestamp is the same as before?

internal/worker/worker.go
```diff
@@ -0,0 +1,221 @@
...
+		WorkerInfo:   worker.info(),
+		RunningTasks: worker.runningTasks(),
+	}
+
+	response, err := worker.rpcClient.Poll(ctx, request, grpc.PerRPCCredentials(worker))
+	if err != nil {
+		return err
+	}
+
+	if response.Shutdown {
```
fkorotkov, line 191: Let's move it right after `TasksToStop` so it's possible to stop and shutdown.

internal/worker/worker_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if sessionTokens[0] != sessionToken {
+		return nil, status.Errorf(codes.PermissionDenied, "invalid session token")
+	}
+
+	return handler(ctx, req)
+}
+
+func TestWorker(t *testing.T) {
```
fkorotkov, line 56: Awesome test! 😍

PERSISTENT-WORKERS.md
```diff
@@ -0,0 +1,56 @@
...
+## Configuration
+
+Path to the YAML configuration can be specified via the `--file` (or `-f` for short version) command-line flag.
+
+Example configuration:
+
+```yaml
+token: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+
+name: "%hostname-%n"
```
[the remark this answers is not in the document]
edigaryev: It will be picked on the server-side, based on the number of current workers connected to a specific pool that share a similar name (expanding all variables except for the `%n` before testing for similarity).

internal/worker/worker.go
edigaryev: I don't think the order should matter here since all started tasks are tied to the `subCtx`, which will be closed once `Run()` exits.

internal/executor/instance/persistentworker_test.go
edigaryev: Added more checks in https://github.com/cirruslabs/cirrus-cli/pull/139/commits/c1fd9ee0f339a499457e5a5fa8a07bffba71cb0a.

fkorotkov (approved):

PR #157, opened 2020-11-10
Externalize podmanapi

edigaryev:
It does live here now: https://github.com/cirruslabs/podmanapi.


fkorotkov (commented):
Let's also add `.cirrus.yml` to https://github.com/cirruslabs/podmanapi.

BTW do you think it will be possible to create a CLI only task for generating the swagger clinet and run it like `cirrus run generate --dirty`?

fkorotkov (approved):

edigaryev:
> Let's also add `.cirrus.yml` to https://github.com/cirruslabs/podmanapi.

I'm not exactly sure which tasks we're going to run there.

> BTW do you think it will be possible to create a CLI only task for generating the swagger clinet and run it like `cirrus run generate --dirty`?

I think this is out-of-scope for the CLI.

fkorotkov:
Just to check it compiles. You mention in the ReadMe how to fix compilation, will be good to catch it in CI if someone will forget to do it. 🤔

PR #172, opened 2020-11-18
Fixed generation of empty list

fkorotkov:
For some reason `yaml.Marshal` of an empty map resulted in `{}` instead of an empty string.

### DIFF
--- a/pkg/larker/larker_test.go
+++ b/pkg/larker/larker_test.go
@@ -26,7 +26,8 @@ func TestSugarCoatedTask(t *testing.T) {
 	validateExpected(t, "testdata/sugar-coated-task")
 }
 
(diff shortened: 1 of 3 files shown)


pkg/larker/larker_test.go
```diff
@@ -26,6 +26,11 @@ func TestSugarCoatedTask(t *testing.T) {
 	validateExpected(t, "testdata/sugar-coated-task")
 }
 
+// TestNoTasks ensures that .cirrus.star that return empty list still generates valid YAML.
```
edigaryev, line 29:
```suggestion
// TestNoTasks ensures that .cirrus.star that return empty list still generates a YAML
// that can be joined with another YAML using only string concatenation.
```

edigaryev (approved):

PR #175, opened 2020-11-19
Build for Apple Silicon

### LINKED ISSUE #171: Release binaries for Apple Silicon
Seems go is not going to support M1 until 1.16 but it [seems it's possible to cross compile arm binary](https://github.com/golang/go/issues/38485#issuecomment-729296273). We should try to build the CLI and the agent binary this way.

fkorotkov:
Via gotip

Will test binary on Friday when the new Mac Mini should arrive.

Related issues: https://github.com/golang/go/issues/42684

Fixes #171

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -87,7 +87,7 @@ task:
 
 task:
   name: Release Apple Silicon
-  only_if: $CIRRUS_TAG != ''
+  only_if: $CIRRUS_RELEASE != ''
   depends_on: Release
   env:
     GITHUB_TOKEN: ENCRYPTED[!98ace8259c6024da912c14d5a3c5c6aac186890a8d4819fad78f3e0c41a4e0cd3a2537dd6e91493952fb056fa434be7c!]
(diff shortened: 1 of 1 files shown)


edigaryev:
These two tasks look very similar, can we use [YAML aliases](https://cirrus-ci.org/guide/tips-and-tricks/#sharing-configuration-between-tasks) instead to reduce the duplication?

fkorotkov:
Though about it but since these tasks are temporary until Goreleaser support Apple Silicon releases I decided not to optimize.  There are slightly different build scripts so only `env`, `install_script` and `build_script` can be refactored in an alias.

.cirrus.yml
```diff
@@ -41,21 +41,74 @@ docker_builder:
...
     image: goreleaser/goreleaser:latest
   release_script: goreleaser build --snapshot
   binaries_artifacts:
     path: "dist/cirrus_*/cirrus*"
 
+task:
+  name: Release Apple Silicon (Dry Run)
+  only_if: $CIRRUS_TAG == ''
+  env:
+    GITHUB_TOKEN: ENCRYPTED[!98ace8259c6024da912c14d5a3c5c6aac186890a8d4819fad78f3e0c41a4e0cd3a2537dd6e91493952fb056fa434be7c!]
```
edigaryev, line 55: This doesn't seem to be required for the dry-run task.

edigaryev:
> There are slightly different build scripts

Why not make them similar? It seems that the only difference should be that we don't upload the results.

.cirrus.yml
fkorotkov: Fixed in b8db09f6caab7badf9989d71c4caff2cfe66f750

fkorotkov:
Refactored out the parts in b8db09f6caab7badf9989d71c4caff2cfe66f750. Also bad news, Mac Mini will arrive only on Tuesday. 😪

edigaryev (approved):

fkorotkov:
It works! 
image

Issue #174, opened 2020-11-19
Config was parsed correctly but no tasks were found!


AlexWayfer:
https://github.com/AlexWayfer/flame-cli/pull/57

Probably because of https://github.com/cirruslabs/cirrus-ci-docs/issues/689

edigaryev:
Please take a look at the CLI's version [0.19.0](https://github.com/cirruslabs/cirrus-cli/releases/tag/v0.19.0) that was just released, it should support `.cirrus.yaml` now.

AlexWayfer:
Sorry, but I can't confirm:

```
> sudo cirrus --version
cirrus version 0.20.1-unknown

> sudo cirrus run
2020/12/03 01:31:01 Config was parsed correctly but no tasks were found!
Error: run failed
2020/12/03 01:31:01 run failed

> ls -lah .cirrus.yaml
-rw-r--r-- 1 alex alex 3.1K Dec  3 01:28 .cirrus.yaml
```

edigaryev:
> Sorry, but I can't confirm:
> 
> ```
> > sudo cirrus --version
> cirrus version 0.20.1-unknown
> 
> > sudo cirrus run
> 2020/12/03 01:31:01 Config was parsed correctly but no tasks were found!
> Error: run failed
> 2020/12/03 01:31:01 run failed
> 
> > ls -lah .cirrus.yaml
> -rw-r--r-- 1 alex alex 3.1K Dec  3 01:28 .cirrus.yaml
> ```

Looks like you have [some `only_if` conditions](https://github.com/AlexWayfer/flame-cli/blob/e49dc0cb4fb93b2ec6ad884080ca9c175a4bece6/.cirrus.yaml#L24-L27) in your `.cirrus.yaml` that don't get evaluated to `true`. This results in parser skipping those tasks.

You can try supplying the missing environment variables to make this work:

```
cirrus run -e CIRRUS_BRANCH=master
```

AlexWayfer:
> > Sorry, but I can't confirm:
> > ```
> > > sudo cirrus --version
> > cirrus version 0.20.1-unknown
> > 
> > > sudo cirrus run
> > 2020/12/03 01:31:01 Config was parsed correctly but no tasks were found!
> > Error: run failed
> > 2020/12/03 01:31:01 run failed
> > 
> > > ls -lah .cirrus.yaml
> > -rw-r--r-- 1 alex alex 3.1K Dec  3 01:28 .cirrus.yaml
[...]

PR #176, opened 2020-11-20
Support .cirrus.yaml too

edigaryev:
See https://github.com/cirruslabs/cirrus-cli/issues/174.

### DIFF
--- a/internal/commands/helpers/helpers.go
+++ b/internal/commands/helpers/helpers.go
@@ -2,6 +2,8 @@ package helpers
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker"
 	"github.com/cirruslabs/cirrus-cli/pkg/larker/fs/local"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


internal/commands/helpers/helpers.go
```diff
@@ -82,18 +83,24 @@ func EvaluateStarlarkConfig(ctx context.Context, path string, env map[string]str
...
+		yamlConfig, yamlErr = ReadYAMLConfig(".cirrus.yml")
 	}
 
-	starlarkConfig, err := EvaluateStarlarkConfig(ctx, ".cirrus.star", env)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return yamlConfig, nil
-		}
-		return "", err
+	starlarkConfig, starlarkErr := EvaluateStarlarkConfig(ctx, ".cirrus.star", env)
```
fkorotkov, line 93: Do you still need to check for `os.IsNotExist(err)`?

edigaryev: Indeed! In the current state it's possible to miss a `cirrus.yml` that fails to read due to e.g. invalid permissions.

edigaryev: Fixed in https://github.com/cirruslabs/cirrus-cli/pull/176/commits/5a59107bdb995f0cc6d28dba4253526771ca79ac.

fkorotkov (approved):

PR #179, opened 2020-11-24
Parser and executor: support Docker Builder

### LINKED ISSUE #178: Support Docker Builder Instances
Should be pretty similar to persistent worker tasks. Instead of Docker container just shell out agent execution locally.

edigaryev:
@fkorotkov this needs metadata properties default value change for the tests to pass.

Resolves #178.

### DIFF
--- a/internal/executor/instance/instance.go
+++ b/internal/executor/instance/instance.go
@@ -20,6 +20,7 @@ import (
 )
 
 var (
+	ErrFailedToCreateInstance    = errors.New("failed to create instance")
 	ErrUnsupportedInstance       = errors.New("unsupported instance type")
 	ErrAdditionalContainerFailed = errors.New("additional container failed")
 )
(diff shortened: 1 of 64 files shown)


internal/executor/instance/instance.go
```diff
@@ -73,6 +74,15 @@ func NewFromProto(anyInstance *any.Any, commands []*api.Command) (Instance, erro
 			Arguments:  instance.Arguments,
 		}, nil
 	case *api.PersistentWorkerInstance:
+		return NewPersistentWorkerInstance()
+	case *api.DockerBuilder:
+		// Ensure that we're not trying to run e.g. Windows-specific scripts on macOS
+		instanceOS := strings.ToLower(instance.Platform.String())
+		if runtime.GOOS != instanceOS {
```
fkorotkov, line 81: you can run linux docker builder on macOS, don't you?

fkorotkov (commented):
Docker builder is essentially a task that has a predefined instance of type DockerBuilder. Do you think `task.Task` can be reused here?

internal/executor/instance/instance.go
edigaryev: I don't think this is a great idea, because this might result in a very surprising behavior. But should it?

Besides, the Docker Builder itself might be somewhat surprising: imagine scripts like `rm -rf /tmp` (or more important paths) will run on the host system.

edigaryev:
> Docker builder is essentially a task that has a predefined instance of type DockerBuilder. Do you think `task.Task` can be reused here?

Implemented in https://github.com/cirruslabs/cirrus-cli/pull/179/commits/1bc73413b9d746fea3c71df9249778f8f154ed12.

fkorotkov (approved):

PR #182, opened 2020-11-29
Image override for Dockerfile as CI environment feature

### LINKED ISSUE #30: Option to override default registry
Right now [Docker as a CI environment](https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment) always uses `gcr.io` as registry. For local runs with CLI it might be usefully to override it. Consider the following use case: CLI is used in a private on-premise Jenkins cluster where there is a local Docker registry under `hub.mycorp.internal`. It will be great to override `gcr.io` with `hub.mycorp.internal` as well as allow to push a newly built `PrebuiltImage` to the registry directly from the CLI.

edigaryev:
Resolves #30.


fkorotkov (approved):
Have you had a chance to see how it works locally with Docker Hub or GitHub Container Registry?

edigaryev:
> Have you had a chance to see how it works locally with Docker Hub or GitHub Container Registry?

Yes, it works, but only for Docker backend. I'll check Podman now.

I've tried to write a test using [github.com/docker/distribution/registry](https://pkg.go.dev/github.com/docker/distribution@v2.7.1+incompatible/registry#NewRegistry), but it just becomes infeasible because you need to re-configure the Docker daemon with `insecure-registries` setting and restart it.

edigaryev:
Got Podman working in https://github.com/cirruslabs/cirrus-cli/pull/182/commits/75725b949cfb1275770a21fea8600e4955d0e8bb.

PR #183, opened 2020-11-30
Include secret in task identity

fkorotkov:
Plus don't fail execution if failed to notify server


edigaryev:
> Plus don't fail execution if failed to notify server

Why not? The server is expected to re-schedule the task in this case, and retry is achieved this way.

fkorotkov:
Retry will be achieved in case of making stop call before task in a final state which will mean the agent exited unexpectedly. Anyways I don't mind reverting it.

edigaryev (approved):

PR #196, opened 2020-12-14
Default worker's name to hostname

fkorotkov:


PERSISTENT-WORKERS.md
```diff
@@ -37,20 +33,22 @@ Example configuration:
...
 ## Writing tasks
 
 Here's an example of how to run a task on one of the persistent workers [registered in the dashboard](https://cirrus-ci.com/):
 
 ```yaml
 task:
-  persistent_worker: {}
+  persistent_worker:
+    labels:
+      os: darwin
```
edigaryev, line 52: This label is reserved:

https://github.com/cirruslabs/cirrus-cli/blob/8abb5935410c7712f8204d31ff9a00879041ec96/internal/worker/worker.go#L96

Perhaps `device` can be used here too, similarly to the example above.

PERSISTENT-WORKERS.md
```diff
@@ -8,25 +8,21 @@ Follow the instruction in the [INSTALL.md](https://github.com/cirruslabs/cirrus-
...
-The most simplest invocation looks like this:
+The simplest invocation looks like this:
 
 ```
 cirrus worker run --token <poll registration token>
 ```
 
 This will start the persistent worker that periodically poll for new tasks in the foreground mode.
 
-By default the worker's name gets assigned semi-automatically based on the name of the current system and a counter variable. Specify `--name` to explicitly specify the worker's name:
```
edigaryev, line 19: The `hostname` seems to be a bit ambiguous here: is it the command output, or a `hostname` string that will be used?

Maybe keep this "name of the current system"?

PERSISTENT-WORKERS.md
fkorotkov: What do you mean? By `device` label I meant a user label for a device that is connected to the mac-mini. Can rename to `connected-device: iPhone12ProMax`

edigaryev: Oh, I see. This `label` thing matches a label instead of setting additional label for the worker to use.

PERSISTENT-WORKERS.md
fkorotkov: See 39eb90ab15c08805013e1176803b44a9f4ea1669

edigaryev (approved):

PR #211, opened 2020-12-29
Parallels isolation support

### LINKED ISSUE #198: Support isolation via Parallels VMs in persistent worker mode
Right now persistent worker mode simple shells out the task without any isolation. Will be great to have an option to specify isolation using Parallels Desktop Pro. For example,

```yaml
mac_m1_task:
  persistent_worker:
    isolation:
      parallels:
        vm_name: big-sur-base
        platform: darwin
        user: admin # for SSHing to start a task
        password: admin # same ^^
    labels:
      os: darwin
      arch: arm64
```

edigaryev:
Resolves #198.


internal/executor/instance/persistentworker/isolation/parallels/vm.go
```diff
@@ -0,0 +1,111 @@
...
+	for _, vmInfo := range vmInfos {
+		if vmInfo.Name == vm.name {
+			return &vmInfo, nil
+		}
+	}
+
+	return nil, fmt.Errorf("%w: failed to find VM %q", ErrVMFailed, vm.name)
+}
+
+func (vm *VM) RetrieveIP(ctx context.Context) (string, error) {
```
fkorotkov, line 88: What do you think about using `prlctl` to get the IP address similar to [`prlctl-ssh`](https://github.com/Starefossen/prlctl-ssh/blob/master/prlctl-ssh.sh#L19-L22)?

edigaryev: This requires guest additions, which we're trying to avoid, as far as I understand.

internal/executor/instance/persistentworker/isolation/parallels/vm.go
```diff
@@ -0,0 +1,119 @@
...
+
+type VirtualMachineInfo struct {
+	ID       string
+	Name     string
+	Hardware HardwareInfo
+}
+
+func NewVMClonedFrom(ctx context.Context, vmNameFrom string) (*VM, error) {
+	vm := &VM{
+		name: "cirrus-" + uuid.New().String(),
```
fkorotkov, line 35: Maybe use task id so the error message will be self explanatory and not confusing?

internal/executor/instance/persistentworker/isolation/parallels/vm.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	_, stderr, err := Prlctl(ctx, "clone", vmNameFrom, "--name", vm.name)
```
fkorotkov, line 38: What do you think about using a linked clone here? It seems it's pretty instantaneous? Not sure about IO implications of this copy-on-write like thing though...
```suggestion
	_, stderr, err := Prlctl(ctx, "clone", vmNameFrom, "--linked", "--name", vm.name)
```

internal/executor/instance/persistentworker/isolation/parallels/vm.go
```diff
@@ -0,0 +1,119 @@
...
+	_, stderr, err = Prlctl(ctx, "delete", vm.name)
+	if err != nil {
+		return fmt.Errorf("%w: failed to delete VM %q: %q", ErrVMFailed, vm.name, firstLine(stderr))
+	}
+
+	return nil
+}
+
+func (vm *VM) retrieveInfo(ctx context.Context) (*VirtualMachineInfo, error) {
+	stdout, stderr, err := Prlctl(ctx, "list", "--info", "--json")
```
fkorotkov, line 76: You can pass `vm.name` here:

```suggestion
	stdout, stderr, err := Prlctl(ctx, "list", "--info", "--json", vm.name)
```

Just a small optimization

internal/executor/instance/persistentworker/isolation/parallels/vm.go
edigaryev: It's already there:

https://github.com/cirruslabs/cirrus-cli/blob/ed3e710a768ecb1df15041007a111923f72179f2/internal/worker/task.go#L50

fkorotkov: But error from `prlctl` are propagated to user and these errors contain this unique VM name without the task id.

fkorotkov (approved):

PR #221, opened 2021-01-07
Expand mapping values with environment variables

fkorotkov:
Related to https://github.com/cirruslabs/cirrus-ci-docs/issues/692


fkorotkov:
Reverted in df366c6ba6e07f1b68e626828ad368f8b7d0e566. It appeared some folks pass encrypted variables as arguments (which is a security risk since docker image will contain the variable unencrypted).

real-or-random:
> Reverted in [df366c6](https://github.com/cirruslabs/cirrus-cli/commit/df366c6ba6e07f1b68e626828ad368f8b7d0e566). It appeared some folks pass encrypted variables as arguments (which is a security risk since docker image will contain the variable unencrypted).

Makes sense. In this case, https://github.com/cirruslabs/cirrus-ci-docs/issues/544 is the simplest change that will help people when they're having issues with the cache.

PR #235, opened 2021-01-15
Boolevator: pre-expand environment variables

edigaryev:
For compatibility with the Cirrus Cloud parser.

### DIFF
--- a/pkg/parser/boolevator/boolevator.go
+++ b/pkg/parser/boolevator/boolevator.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/PaesslerAG/gval"
+	"github.com/cirruslabs/cirrus-cli/pkg/parser/expander"
 	"strconv"
 	"strings"
 	"text/scanner"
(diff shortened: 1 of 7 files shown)


pkg/parser/boolevator/boolevator.go
```diff
@@ -44,6 +45,11 @@ func (boolevator *Boolevator) Eval(expr string, env map[string]string) (bool, er
 	// Work around text/scanner stopping at newline when scanning strings
 	expr = strings.ReplaceAll(expr, "\n", "\\n")
 
+	// Pre-expand environment variables
+	for key, value := range env {
+		env[key] = expander.ExpandEnvironmentVariables(value, env)
```
fkorotkov, line 50: Let's create a new map so the `Eval` won't have side effects.

fkorotkov (changes requested):

pkg/parser/boolevator/boolevator.go
edigaryev: Good catch! Fixed in e519dd374153223383894d50b8415f2b287dabea.

pkg/parser/boolevator/boolevator.go
```diff
@@ -45,9 +45,15 @@ func (boolevator *Boolevator) Eval(expr string, env map[string]string) (bool, er
...
+	// Ensure that we keep the env as is
+	localEnv := make(map[string]string)
 	for key, value := range env {
-		env[key] = expander.ExpandEnvironmentVariables(value, env)
+		localEnv[key] = value
+	}
+
+	// Pre-expand environment variables
+	for key, value := range localEnv {
+		localEnv[key] = expander.ExpandEnvironmentVariables(value, localEnv)
```
fkorotkov, line 56: Can it be replaced with?

```suggestion
		localEnv[key] = expander.ExpandEnvironmentVariables(value, env)
```

fkorotkov (approved):

PR #246, opened 2021-01-19
[worker] force time sync for Parallels VMs

fkorotkov:
Suspended VMs don't pick time right away


fkorotkov:
One option can be to put the same command to `~/.profile` but I think an extra sync won't hurt. Also the current Anka integration is doing this.

edigaryev (approved):

Issue #253, opened 2021-01-21
Support other platforms for Parallels isolation


fkorotkov:
Right now only `darwin` platform is supported. It will be great to close the loop and support Linux and Windows VMs as well.

edigaryev:
Any idea how we're going to SSH into Windows and also do this (which is somewhat distro-specific because there are many `NTP` implementations)? 🤔 

https://github.com/cirruslabs/cirrus-cli/blob/6ce9f6b2ed22b9df0d4c4c1f27abae9db226efdf/internal/executor/instance/persistentworker/isolation/parallels/parallels.go#L106-L109

fkorotkov:
If Packer builder supports any Windows host then we should check how it's doing it. 🤔 Anyways, created the issue for tracking. It's not a priority but worth investigating once bandwidth allows.

PR #254, opened 2021-01-21
[skip ci] Document isolation options

fkorotkov:


PERSISTENT-WORKERS.md
```diff
@@ -84,3 +84,43 @@ task:
...
+### Isolation
+
+By default, a persistent worker does not isolate execution of a task. All the task instructions are executed directly on
+the worker which can have side effects. This is intended since the main use case for persistent workers is to test on
+bare metal. There are options to enable isolation of task execution:
+
+#### Parallels Desktop for Mac
+
+If your host has [Parallels Desktop](https://www.parallels.com/products/desktop/) installed, then a persistent worker
+can execute tasks in available Parallels VMs (worker will clone a VM, run the task and then remote the temporary cloned
```
edigaryev, line 97:
```suggestion
can execute tasks in available Parallels VMs (worker will clone a VM, run the task and then remove the temporary cloned
```

edigaryev (approved):
I think this is somewhat too early to document this due known unknowns in #253 which might result in YAML syntax change, but otherwise LGTM, left a small nit.

edigaryev (approved):
I think this is somewhat too early to document this due known unknowns in #253 which might result in YAML syntax change, but otherwise LGTM, left a small nit.

PR #255, opened 2021-01-22
Use built-in Go parser by default

### LINKED ISSUE #251: Switch to Go parser by default
I think it's time to switch to Go parser by default instead of the RPC one. The Go parser is almost there are is working perfectly in most of the cases. IMO we can switch the default parser for the CLI while still polishing last corner cases of compatibility with the cloud parser.

edigaryev:
The old parser can be enabled with `--experimental-old-parser`.

Resolves #251.

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -41,7 +41,7 @@ var dockerfileImagePush bool
 var debugNoCleanup bool
 
 // Experimental features flags.
-var experimentalParser bool
+var experimentalOldParser bool
 
 func run(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


fkorotkov:
Maybe call the flag `--legacy-remote-parser`?

edigaryev:
> Maybe call the flag `--legacy-remote-parser`?

See https://github.com/cirruslabs/cirrus-cli/pull/255/commits/e4c2a68ab6cd9e4ca286e1f68634a6cd35d84e78.

fkorotkov (approved):

Issue #256, opened 2021-01-25
Failing to prebuild dockerfile: "invalid reference format"


instagibbs:
Cirrus CI jobs are completing for https://github.com/bitcoin-core/HWI/blob/master/.cirrus.yml, but I'm getting stuck running locally:

```
cirrus run
🕕 'Prebuild ci/cirrus.Dockerfile' Task 0.6s
   Pulling image gcr.io/cirrus-ci-community/gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest:latest...
   Image gcr.io/cirrus-ci-community/gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest:latest is not available locally nor remotely, building it...

Error: Error response from daemon: invalid reference format
2021/01/25 13:40:45 Error response from daemon: invalid reference format
```

Ideas?

edigaryev:
Hello Gregory, thanks for your report!

This should be now fixed as a part of [0.30.5 release](https://github.com/cirruslabs/cirrus-cli/releases/tag/v0.30.5), the binaries and the new Homebrew formula will be available shortly.

instagibbs:
efficient, thank you!

instagibbs:
Apologies, but now I'm getting:

```
cirrus run
✅ 'Prebuild ci/cirrus.Dockerfile' Task 25s
🕑 'lint' task 1s
   ✅ Preparing execution environment... 1.3s
Error: Error response from daemon: No such image: gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest
2021/01/26 09:54:32 Error response from daemon: No such image: gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest
```

edigaryev:
>`Error response from daemon: No such image:`

Hmm, it looks that the image built in the `Prebuild ci/cirrus.Dockerfile` step has a different name than `gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest` for some reason or it was deleted before the actual task has started.

Could you please re-run the build using the `cirrus run -o simple -v` command?

instagibbs:
```
cirrus run -o simple -v
gRPC server is listening at http://127.0.0.1:63931 (http://host.docker.internal:63931 inside of a container)
running task Prebuild ci/cirrus.Dockerfile (48)
[...]

PR #261, opened 2021-01-27
Assorted prebuilt instance fixes

edigaryev:
Should help with the issues in #256.

### DIFF
--- a/internal/executor/instance/containerbackend/podman_linux.go
+++ b/internal/executor/instance/containerbackend/podman_linux.go
@@ -240,6 +240,10 @@ func (backend *Podman) ImageBuild(
 		}
 		q.Add("buildargs", string(jsonArgs))
 
+		if input.Pull {
+			q.Add("pull", "true")
+		}
+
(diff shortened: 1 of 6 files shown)


internal/executor/instance/containerbackend/podman_linux.go
```diff
@@ -240,6 +240,10 @@ func (backend *Podman) ImageBuild(
 		}
 		q.Add("buildargs", string(jsonArgs))
 
+		if input.Pull {
```
fkorotkov, line 243: Is it happening for Podman only? Is there a need for Docker equivalent like `--pull` flag?

edigaryev: >Is it happening for Podman only?

It's not: https://github.com/cirruslabs/cirrus-cli/pull/261/files#diff-c581abc4ab75c4bdf8b235d25334d7362af9e1521a53d55a727d0ce84ff8814eR126.

fkorotkov (approved):

PR #263, opened 2021-01-27
Rewrote matrix modifier to work on *node.Node instead of yaml.MapSlice

edigaryev:
This will allow us to more easily abstract from the `yaml.MapSlice` and freely use any other YAML package data structure, e.g. [yaml.v3's Node](https://pkg.go.dev/gopkg.in/yaml.v3#Node), which contains uninterpreted scalar values as strings and line/column numbers.


pkg/parser/modifier/matrix/matrix.go
```diff
@@ -14,84 +14,60 @@ var ErrMatrixNeedsCollection = errors.New("matrix should contain a collection")
...
+func singlePass(inputTree *node.Node) error {
+	// Find a matrix node to expand
+	matrixNode := inputTree.DeepFindChild("matrix")
+	if matrixNode == nil {
+		return errNoExpansionDone
+	}
 
-	if len(inputTree) == 0 {
-		return nil, errNoExpansionDone
+	// Ensure this matrix node is attached to either a task or a docker_builder
```
fkorotkov, line 31: I wonder if it should work with `pipe` 🤔

pkg/parser/node/matrix.go
```diff
@@ -0,0 +1,106 @@
...
+			break
+		}
+
+		parent = parent.Parent
+	}
+
+	return result
+}
+
+func (node *Node) GetPath(path []int) *Node {
```
fkorotkov, line 58: Maybe make a type alias for `TreePath`s instead of using `[]int`?

pkg/parser/modifier/matrix/matrix.go
edigaryev: Should it?

pkg/parser/node/matrix.go
edigaryev: I don't think it'd make any difference, because virtually all path operations need to know that this is a slice of ints and on top of that all of this is only used internally.

fkorotkov (approved):

PR #265, opened 2021-01-29
Use GWDG's mirror instead of broken openSUSE traffic director

edigaryev:
Should help with Podman installation failing.


fkorotkov (approved):

edigaryev:
Found here: https://mirrors.opensuse.org/list/repositories.html.

PR #274, opened 2021-02-03
Abstract yaml internals

fkorotkov:
Moved all production use of `yaml.v3` package to `pkg/parser`

And specifically `yaml.go` which has all the logic for parsing YAML and printing out pretty presentation of Node.


edigaryev (commented):
What about `internal/commands/internal/test/test.go` and `internal/commands/internal/test/configuration.go`?

But I generally think that making `larker` use the `parser.Node` is a bad idea, because this imposes unrelated guarantees on `Node` that it might not hold in the future.

Making non-`parser` packages depend on `Node` in a task where only basic YAML constructs are needed to get the job done seems strange. Why do that?

If abstraction desired, I'd rather suggest implementing even our own `yaml.MapItem` (should get `yaml.MapSlice` for free) for `yaml.v3`, because it will additionally would allow for easier migration in case we'd need it. See https://github.com/go-yaml/yaml/issues/698.

fkorotkov:
@edigaryev how about these parts then https://github.com/cirruslabs/cirrus-cli/pull/275? It keeps larker use `yaml.v3` but moves yaml helpers for a separate package so `larker` doesn't depend on `node`.

PR #275, opened 2021-02-03
Restructure yaml

fkorotkov:
Some structural changes and moved around code instead of having `larker` depend on `node`.


edigaryev (approved):

pkg/helpers/yaml/yaml.go
```diff
@@ -1,6 +1,27 @@
-package utils
+package yaml
```
edigaryev, line 1: Also, consider moving this file to `pkg/parser/yamlhelper/yamlhelper.go`, this way:

* you don't need to import it with a different name (see below)
* it will be easier to understand what this package is about (as opposed to common names like "utils")

fkorotkov: Moved to `pkg/yamlhelper`

PR #278, opened 2021-02-05
Parallels: cleanup & error message fixes

edigaryev:
Multiple changes in separate commits for clarity

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/parallels/clone.go
+++ b/internal/executor/instance/persistentworker/isolation/parallels/clone.go
@@ -10,6 +10,9 @@ import (
 func cloneFromSuspended(ctx context.Context, vmPathFrom string) (*VM, error) {
 	vm := &VM{
 		uuid: uuid.New().String(),
+
+		shouldRenewDHCP:  true,
+		delayedIsolation: true,
 	}
 
 	serverInfo, err := GetServerInfo(ctx)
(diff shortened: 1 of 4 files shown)


internal/executor/instance/persistentworker/isolation/parallels/clone.go
```diff
@@ -56,14 +44,5 @@ func cloneFromDefault(ctx context.Context, vmNameFrom string) (*VM, error) {
 		return nil, fmt.Errorf("%w: failed to clone VM %q: %q", ErrVMFailed, vm.Ident(), firstNonEmptyLine(stderr))
 	}
 
-	if err := vm.isolate(ctx); err != nil {
```
fkorotkov, line 59: Should it set `delayedIsolation`?

edigaryev: It's already set here: https://github.com/cirruslabs/cirrus-cli/pull/278/files#diff-ad16c76192d04ea42699ba6c12657f05dff37e2e2df1972197ab29b93b04305aR16.

fkorotkov: It's `cloneFromDefault` 🤔

edigaryev: Ahh, you don't need to set `delayedIsolation` for `cloneFromDefault` because it's not delayed there (VM is not suspended and the isolation can be set before the VM is started, that's exactly how it's worker before).

fkorotkov (approved):

PR #291, opened 2021-02-16
Stream launched container's logs to the CLI's logger

### LINKED ISSUE #187: Stream launched container's logs to the CLI's logger
This is helpful for the root cause diagnosis, especially when implementing new functionality.

There will be a quite amount of logs, but doing this with the trace level should probably be OK from the UX standpoint.

edigaryev:
Resolves #187.

### DIFF
--- a/internal/executor/instance/containerbackend/podman_linux.go
+++ b/internal/executor/instance/containerbackend/podman_linux.go
@@ -1,6 +1,7 @@
 package containerbackend
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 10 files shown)


internal/executor/instance/containerbackend/podman_linux.go
```diff
@@ -72,6 +76,15 @@ func NewPodman() (ContainerBackend, error) {
 		HTTPClient: podman.httpClient,
 	})
 
+	// Query server's version and activate bug workarounds (if applicable)
+	version, err := podman.SystemInfo(context.Background())
+	if err != nil {
+		return nil, err
+	}
+	if version.Version == "3.0.0" {
```
fkorotkov, line 84: Should it check if it's 3.0.0+?

edigaryev: I hope this will get fixed soon, so only checking for the broken version here.

fkorotkov (approved):

PR #295, opened 2021-02-16
Remove --container-pull and introduce --container-lazy-pull

### LINKED ISSUE #290: #181 causes bad ux
#181 causes bad UX because the default configuration differs in behavior from the cloud version, and the resulting errors may not be obvious as to why.

Yeah, having a user-facing knob for this is good and useful. But given that there's no time-based limits on image checking, pulling every run is reasonable. (At least, once per run if the image is used in multiple places.)

edigaryev:
Resolves #290.


internal/commands/run.go
```diff
@@ -158,8 +158,8 @@ func newRunCmd() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&containerBackend, "container-backend", containerbackend.BackendAuto,
 		fmt.Sprintf("container engine backend to use, either \"%s\", \"%s\" or \"%s\"",
 			containerbackend.BackendDocker, containerbackend.BackendPodman, containerbackend.BackendAuto))
-	cmd.PersistentFlags().BoolVar(&containerPull, "container-pull", false,
-		"force pull the images before starting containers")
+	cmd.PersistentFlags().BoolVar(&containerLazyPull, "container-lazy-pull", false,
+		"when starting containers only pull images that are not available locally")
```
fkorotkov, line 162: Maybe

```suggestion
		"attempt to pull images only if they are missing locally (helpful in case of registry rate limits)")
```

To highlight there otherwise there will be not even an attempt to pull?

internal/commands/run_test.go
```diff
@@ -221,14 +221,14 @@ func TestRunContainerPull(t *testing.T) {
...
-	testutil.TempChdirPopulatedWith(t, "testdata/docker-no-pull")
+	testutil.TempChdirPopulatedWith(t, "testdata/image-pulling-behavior")
 
 	// Create os.Stderr writer that duplicates it's output to buf
 	buf := bytes.NewBufferString("")
 	writer := io.MultiWriter(os.Stderr, buf)
 
 	command := commands.NewRootCmd()
-	command.SetArgs([]string{"run", "-v", "-o simple", "--container-pull"})
+	command.SetArgs([]string{"run", "-v", "-o simple"})
```
fkorotkov, line 231: Should we put `--container-lazy-pull` to all other `run` invocations in test so the tests are graceful on the Docker Hub limits?

edigaryev: It's already working this way, see `EagerPull` field in `ContainerOptions`.

fkorotkov: I believe `EagerPull` is set to `true` in tests because `containerLazyPull` is `false` by default and `EagerPull` is `!containerLazyPull`.

edigaryev: Oh indeed, I was thinking of all other tests that don't use the `run` command.

edigaryev: Fixed in https://github.com/cirruslabs/cirrus-cli/pull/295/commits/e578304884f6dd2d1b27aa0586a6f8d0b038d6df.

fkorotkov (approved):

PR #297, opened 2021-02-17
No need for gotip anymore

edigaryev:
Go 1.16 was just released and GoReleaser introduced darwin/arm64 support [a few hours ago](https://github.com/goreleaser/goreleaser/releases/tag/v0.156.0).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -74,29 +74,6 @@ task:
   binaries_artifacts:
     path: "dist/cirrus_*/cirrus*"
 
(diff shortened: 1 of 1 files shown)


edigaryev:
It seems that the `golang:latest` was cached and a version previous to `1.16` is used, which results in a linker error.

fkorotkov:
It seems it took a while to push the new image. Seems they updated just an hour ago.

fkorotkov (approved):

fkorotkov:
Should probably update .cirrus.umlaut to use go 1.16

PR #302, opened 2021-02-18
Don't assume platform by default

fkorotkov:
If proto instance like `persistent_worker` doesn't have `platform` field then don't assume it's `linux` and simply don't set `CIRRUS_OS`.

### DIFF
--- a/internal/evaluator/evaluator_test.go
+++ b/internal/evaluator/evaluator_test.go
@@ -205,6 +205,21 @@ proto_task:
 	})
 }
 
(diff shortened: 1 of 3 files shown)


edigaryev:
A test would be nice for this one.

fkorotkov:
I tested it with the Clouds integration tests where it was failing. Let me try to find a test case for instances that CLI has. 🤔

fkorotkov:
Added a test which is failing on `master`. See 4d4e0db0babd18872025474e5c34c7b4ad1a4bac

edigaryev (approved):

PR #303, opened 2021-02-19
Starlark hooks support

### LINKED ISSUE #143: Starlark hooks
Since we're now supporting Starlark, some build-time decisions can be handled there instead of writing the logic in the YAML.

There are a couple use-cases that we can cover by letting users to define their own hooks:

* customizing re-run logic: https://github.com/cirruslabs/cirrus-ci-docs/issues/238
* handling GitHub events: https://github.com/cirruslabs/cirrus-cli/issues/45#issuecomment-644872742

The hook name for the first use-case will be probably defined explicitly in the YAML to avoid limiting all tasks to a single hook, while the latter will be defined implicitly based on whether a function with a specific name exists in `.cirrus.star` or not.

edigaryev:
Resolves #143.


pkg/larker/larker.go
```diff
@@ -112,3 +124,98 @@ func (larker *Larker) Main(ctx context.Context, source string) (string, error) {
...
+	ctx context.Context,
+	source string,
+	name string,
+	arguments []interface{},
+) (*HookResult, error) {
+	if name == "" {
+		return nil, fmt.Errorf("%w: empty hook name specified", ErrSanity)
+	}
+
+	if len(arguments) != 1 {
```
fkorotkov, line 138: It shouldn't matter. We'll decide it on the server side.

pkg/larker/larker.go
```diff
@@ -112,3 +124,98 @@ func (larker *Larker) Main(ctx context.Context, source string) (string, error) {
...
+			Result:        hookResultStarlark,
+		}
+	}()
+
+	select {
+	case hookResult := <-resCh:
+		return hookResult, nil
+	case err := <-errCh:
+		return &HookResult{ErrorMessage: err.Error()}, err
+	case <-ctx.Done():
```
fkorotkov, line 218: Should we do something about `thread` here? Like calling `thread.Cancel("canceled")`

pkg/larker/larker.go
```diff
@@ -112,3 +124,98 @@ func (larker *Larker) Main(ctx context.Context, source string) (string, error) {
...
+		return nil, fmt.Errorf("%w: empty hook name specified", ErrSanity)
+	}
+
+	if len(arguments) != 1 {
+		return nil, fmt.Errorf("%w: hook takes exactly 1 argument, got %d", ErrSanity, len(arguments))
+	}
+
+	outputLogsBuffer := &bytes.Buffer{}
+
+	thread := &starlark.Thread{
```
fkorotkov, line 144: I think we need to set some meaningful `SetMaxExecutionSteps` here.

pkg/larker/larker.go
edigaryev: Good catch, fixed in https://github.com/cirruslabs/cirrus-cli/pull/303/commits/2711b422c22e4bc2ca140c06f4665b59c2df98c1.

pkg/larker/larker.go
edigaryev: Shouldn't https://github.com/cirruslabs/cirrus-cli/pull/303/commits/2711b422c22e4bc2ca140c06f4665b59c2df98c1 cover this already?

pkg/larker/larker.go
edigaryev: Changed in https://github.com/cirruslabs/cirrus-cli/pull/303/commits/5b990fe1f4a554bee975f65c2a7e8331da170b77.

fkorotkov (approved):

PR #304, opened 2021-02-22
cirrus validate: query additional instances from the RPC

### LINKED ISSUE #288: Fix "task has no instance attached" for locally unsupported instances
Right now it seems we check for instance being set right after parsing but we should validate it after filtration because some instance types are not supported but it shouldn't affect running of a single task that in fact is supported.

To reproduce simply run `cirrus run lint` in this repository and you'll get:

```bash
❯ cirrus run --debug-no-cleanup lint 
86:     - gotip version
87:   build_script:
88:     - CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 gotip build -o cirrus-darwin-arm64 -ldflags "-X github.com/cirruslabs/cirrus-cli/internal/version.Version=$CIRRUS_TAG" cmd/cirrus/main.go
89: 
90: task:
91:   name: Release Apple Silicon (Dry Run)
      ^
92:   only_if: $CIRRUS_TAG == ''
93:   << : *APPLE_SILICON_BUILD
94:   binary_artifacts:
95:     path: cirrus-darwin-arm64
96: 
Error: parsing error: 91:3: task has no instance attached
```

edigaryev:
Resolves #288.

### DIFF
--- a/internal/commands/validate/validate.go
+++ b/internal/commands/validate/validate.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	"github.com/cirruslabs/cirrus-ci-agent/api"
 	"github.com/cirruslabs/cirrus-cli/internal/commands/helpers"
+	"github.com/cirruslabs/cirrus-cli/internal/evaluator"
 	eenvironment "github.com/cirruslabs/cirrus-cli/internal/executor/environment"
 	"github.com/cirruslabs/cirrus-cli/internal/testutil"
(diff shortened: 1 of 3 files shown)


fkorotkov (approved):

edigaryev:
> Maybe add a test that has `container` and `freebsd_instance` tasks?

Added a test in https://github.com/cirruslabs/cirrus-cli/pull/318.

PR #339, opened 2021-03-10
Evaluator: send output_logs

edigaryev:
`github.com/cirruslabs/cirrus-ci-agent`'s version can be changed to a proper one after https://github.com/cirruslabs/cirrus-ci-agent/pull/111 gets merged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/antihax/optional v1.0.0
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/certifi/gocertifi v0.0.0-20200922220541-2c3bb06c6054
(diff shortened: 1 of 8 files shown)


fkorotkov (commented):
Can we add a test that have two print statments and verify that the output bytes will have a new line between them?

```python
print("Foo")
print("Bar")
```

Will result in `Foo\nBar\n`?

edigaryev:
> Can we add a test that have two print statments and verify that the output bytes will have a new line between them?
> 
> ```python
> print("Foo")
> print("Bar")
> ```
> 
> Will result in `Foo\nBar\n`?

Added a test in 8603f92.

fkorotkov (approved):

PR #344, opened 2021-03-18
Parser: check required fields

edigaryev:


pkg/parser/parseable/parseable.go
```diff
@@ -51,17 +49,30 @@ func (parser *DefaultParser) Parse(node *node.Node) error {
 		}
 	}
 
+	seenFields := map[string]struct{}{}
+
 	for _, child := range node.Children {
 		for _, field := range parser.fields {
 			if field.name.Matches(child.Name) {
+				seenFields[field.name.MapKey()] = struct{}{}
```
fkorotkov, line 57: why use `map[string]struct{}{}` and not something like `map[string]bool{}`? This seems just allocating object for nothing. It's not a concern but I'm just curious 😅

edigaryev: That's just a more semantically sound way of saying that map values are unused, compared to `bool` (which has two states).

A little more in depth take on this [in /r/golang's comment](https://www.reddit.com/r/golang/comments/gnge5m/stop_using_mapstringstruct_for_existence_checks/fr9udzk/).

fkorotkov (approved):

PR #345, opened 2021-03-24
Support verifying logs while template testing

fkorotkov:
Should be handy for templates that doesn't generate tasks.


internal/commands/internal/test/test.go
```diff
@@ -96,27 +97,43 @@ func test(cmd *cobra.Command, args []string) error {
...
-	logger.Finish(someTestsFailed)
+	logger.Finish(!someTestsFailed)
 	if someTestsFailed {
 		return fmt.Errorf("%w: some tests failed", ErrTest)
 	}
 
 	return nil
 }
 
+func logDifferenceIfAny(logger *echelon.Logger, a, b interface{}) bool {
```
edigaryev, line 127: There needs to be a way to differentiate between differences somehow, since there might be a test that needs to check both the config and logs for differences.

The output for a such test would look something like this:

```
❌ internal/commands/internal/test/testdata/logs 0.0s
   Kind: 1 != 0
   Content: [0xc000513c20] != <nil slice>
   Line: 1 != 0
   Column: 1 != 0
   slice[1]: 123 != 
   slice[2]:  != <no value>
```

internal/commands/internal/test/test_test.go
```diff
@@ -32,3 +33,23 @@ func TestSimple(t *testing.T) {
 	adaptedPath := filepath.FromSlash("dir/subdir")
 	assert.Contains(t, buf.String(), fmt.Sprintf("'%s' succeeded", adaptedPath))
 }
+
+func TestDryRunAll(t *testing.T) {
```
edigaryev, line 37: This is a bit confusing: any `cirrus internal test` invocation is a dry run.

Why not just test the specific little thing, and abstract away the boilerplate if needed?

fkorotkov: All the tests cases so far are valid and expected to succeed. This test just dry runs all of the them to make sure they succeed without extra checks like other tests (checking logs, etc.).

Can rename to `TestAll` 🤷‍♂️

internal/commands/internal/test/test.go
fkorotkov: What is your suggestion? This weird output seems to be from the initial comparison of raw `Node`s.

edigaryev: >What is your suggestion?

```patch
diff --git a/internal/commands/internal/test/test.go b/internal/commands/internal/test/test.go
index 44ce841..b441707 100644
--- a/internal/commands/internal/test/test.go
+++ b/internal/commands/internal/test/test.go
@@ -97,13 +97,14 @@ func test(cmd *cobra.Command, args []string) error {
                        return fmt.Errorf("%w: %v", ErrTest, err)
                }
 
-               differentConfig := logDifferenceIfAny(logger, expectedConfig, generatedConfig)
+               differentConfig := logDifferenceIfAny("YAML configuration", logger, expectedConfig, generatedConfig)
 
                logsBytes, err := ioutil.ReadFile(filepath.Join(testDir, ".cirrus.expected.log"))
                if err != nil && !errors.Is(err, os.ErrNotExist) {
                        return fmt.Errorf("%w: %v", ErrTest, err)
                }
                differentLogs := logDifferenceIfAny(
+                       "logs",
                        logger,
                        strings.Split(string(logsBytes), "\n"),
                        strings.Split(string(result.OutputLogs), "\n"),
@@ -124,9 +125,11 @@ func test(cmd *cobra.Command, args []string) error {
        return nil
 }
 
-func logDifferenceIfAny(logger *echelon.Logger, a, b interface{}) bool {
+func logDifferenceIfAny(where string, logger *echelon.Logger, a, b interface{}) bool {
        diff := deep.Equal(a, b)
 
+       logger.Warnf("detected difference in %s:", where)
+
        for _, line := range diff {
                logger.Warnf("%s", line)
        }
```

internal/commands/internal/test/test_test.go
edigaryev: That (or at least a comment that clarifies why it's "dry") would be better.

internal/commands/internal/test/test.go
fkorotkov: See 5a6df8777828aee7b801c276eea6e4f6f68337c1

internal/commands/internal/test/test_test.go
fkorotkov: See e756614

edigaryev (approved):

Issue #357, opened 2021-04-05
Persistent worker: remove working directory cleaner


edigaryev:
Added in #356.

Also see https://github.com/cirruslabs/cirrus-cli/issues/352#issuecomment-811976981.

MarcoFalke:
I just tried v0.45.0-07fb74e, but it didn't clear the directory. I presume the cirrus agent needs to be bumped on the cirrus servers first?

edigaryev:
> I just tried v0.45.0-07fb74e, but it didn't clear the directory. I presume the cirrus agent needs to be bumped on the cirrus servers first?

I'm sorry you've had to test this by yourself, there was a logic error in the agent that should be fixed as a part of [0.45.1 CLI's release](https://github.com/cirruslabs/cirrus-cli/releases/tag/v0.45.1) (which itself will point to the latest agent release with the fix), binaries for which will be available shortly 🙌

MarcoFalke:
No worries, thanks for working on this. I just re-tested and it looks like it prefers the directory name `/tmp/cirrus-build-557367079/` over `/tmp/cirrus-ci-build/`

edigaryev:
> No worries, thanks for working on this. I just re-tested and it looks like it prefers the directory name `/tmp/cirrus-build-557367079/` over `/tmp/cirrus-ci-build/`

Thanks, this means it's working. These should also be removed automatically after the task completes.

This issue is just a TODO to remove the logic in https://github.com/cirruslabs/cirrus-cli/pull/356/commits/6004e6fdbf9c4f0d0e456004bc901329f149a078 after some time in the future.

MarcoFalke:
Oh, I wanted to say that if a random directory name is picked (which may be included in cache artefacts), then it will be harder to re-use a cache between runs. We might be able to strip the path name from the cache, but if it is easy to always pick the `/tmp/cirrus-ci-build/` folder, then we'll probably prefer that.

edigaryev:
> Oh, I wanted to say that if a random directory name is picked (which may be included in cache artefacts), then it will be harder to re-use a cache between runs. We might be able to strip the path name from the cache, but if it is easy to always pick the `/tmp/cirrus-ci-build/` folder, then we'll probably prefer that.

[...]

PR #358, opened 2021-04-06
Bump default agent's version to 1.38.1

edigaryev:
See https://github.com/cirruslabs/cirrus-cli/issues/357#issuecomment-814082010.


fkorotkov:
Will do on server side too so it automatically auto-upgrades.

fkorotkov (approved):

PR #360, opened 2021-04-06
Persistent worker: prefer static temporary directory over dynamic one when possible

edigaryev:
To improve caching, see https://github.com/cirruslabs/cirrus-cli/issues/357#issuecomment-814250111.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/none/none.go
+++ b/internal/executor/instance/persistentworker/isolation/none/none.go
@@ -26,12 +26,12 @@ type PersistentWorkerInstance struct {
 
 func staticTempDirWithDynamicFallback() (string, error) {
 	// Prefer static directory for non-Cirrus CI caches efficiency (e.g. ccache)
-	staticTempDir := filepath.Join(os.TempDir(), "cirrus-cli-build")
+	staticTempDir := filepath.Join(os.TempDir(), "cirrus-build")
 	if err := os.Mkdir(staticTempDir, 0700); err == nil {
 		return staticTempDir, nil
 	}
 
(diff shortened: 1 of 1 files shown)


internal/executor/instance/persistentworker/isolation/none/none.go
```diff
@@ -23,9 +24,19 @@ type PersistentWorkerInstance struct {
 	cleanup func() error
 }
 
+func staticTempDirWithDynamicFallback() (string, error) {
+	// Prefer static directory for non-Cirrus CI caches efficiency (e.g. ccache)
+	staticTempDir := filepath.Join(os.TempDir(), "cirrus-cli-build")
```
fkorotkov, line 29: Let's do `cirrus-build` and maybe we'll change it in the cloud too for consistency

```suggestion
	staticTempDir := filepath.Join(os.TempDir(), "cirrus-build")
```

edigaryev: Why? The `ci`/`cli` designation might aid in debugging.

fkorotkov: To possibly reuse caches between worker tasks and regular container tasks (in case caches contain absolute paths). 🤷‍♂️

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/360/commits/2ccc7fa4f1df281eb56272143d0825a06dc21d7c.

fkorotkov (approved):

PR #365, opened 2021-04-13
Parser: only_if-aware dependency validation

### LINKED ISSUE #355: Validate dependecy names
Validate that all task names in `depends_on` are from existing tasks:

```yaml
container:
  image: latest

foo_task:
  script: echo foo

bar_task:
  depends_on: fooo
  script: echo bar
```

edigaryev:
Re-implemented https://github.com/cirruslabs/cirrus-cli/pull/138/commits/a48c6246c39abe95eac09047d1ce22d8009e458b to be `only_if`-aware.

Resolves #355.


pkg/parser/parser.go
```diff
@@ -300,19 +306,34 @@ func (p *Parser) Schema() *schema.Schema {
...
 }
 
-func resolveDependenciesShallow(tasks []task.ParseableTaskLike) {
+func (p *Parser) resolveDependenciesShallow(tasks []task.ParseableTaskLike) error {
 	for _, task := range tasks {
 		var dependsOnIDs []int64
 		for _, dependsOnName := range task.DependsOnNames() {
+			// Dependency may be missing due to only_if
+			if _, ok := p.disabledTaskNamesAndAliases[dependsOnName]; ok {
+				break
```
fkorotkov, line 315: I think it should be `continue`. Seems you don't want to completely break.

fkorotkov: I think an test case like this will be incorrect:

```yaml
container:
  image: latest

foo_task:
  depends_on: 
    - bar
    - baz
  script: echo foo

bar_task:
  only_if: false
  script: echo bar

baz_task:
  script: echo baz
```

edigaryev: Good catch! Fixed in 60923ac.

fkorotkov (approved):

PR #376, opened 2021-05-04
Bump github.com/qri-io/starlib to 0.5.0

edigaryev:
To include https://github.com/qri-io/starlib/pull/71.


fkorotkov (approved):

pkg/larker/loader/loader.go
```diff
@@ -146,11 +146,15 @@ func (loader *Loader) loadCirrusModule() (starlark.StringDict, error) {
...
 
-	jsonModule, err := json.LoadModule()
-	if err != nil {
-		return nil, err
+	// Work around https://github.com/qri-io/starlib/pull/70
+	fixedJSONModule := &starlarkstruct.Module{
+		Name: "json",
+		Members: starlark.StringDict{
+			"loads": starlarkjson.Module.Members["decode"],
+			"dumps": starlarkjson.Module.Members["encode"],
```
fkorotkov, line 154: Should we also migrate to `decode`/`encode`?

edigaryev: But why? The standard way is to use `json.dumps()` and `json.loads()` in Python, and re-using these names should make it more obvious.

PR #378, opened 2021-05-06
Persistent worker: support container isolation

### LINKED ISSUE #374: Support Docker isolation for persistent worker mode
Right now only None or Parallels isolations are supported. In some rare cases it seems it's reasonable to support Docker isolation as well. I know it's possible to just use `container`s for that reason but here is a use case:

Building Android Open Source Project requires a huge `ccache`. It's a waste to always archive and unarchive `ccache`. But is a persistent worker could have a Docker isolation with an option to mount a local `ccache` folder then it should all work perfectly. See:

```yaml
persistent_worker:
  isolation:
    docker:
      image: ubuntu:latest
      cpu: 24
      memory: 60G
      mount:
        - /root/.ccache:/root/.ccache
```

edigaryev:
Resolves #374.


internal/executor/instance/container/run.go
```diff
@@ -0,0 +1,329 @@
...
+			Type:   containerbackend.MountTypeVolume,
+			Source: params.WorkingVolumeName,
+			Target: params.WorkingDirectory,
+		})
+	}
+
+	for _, volume := range params.Volumes {
+		input.Mounts = append(input.Mounts, containerbackend.ContainerMount{
+			Type:     containerbackend.MountTypeBind,
+			Source:   volume.Source,
```
fkorotkov, line 136: Should we check here that the source exists? And create an empty directory if not. Or Docker won't fail on a missing folder 🤔

internal/executor/executor_test.go
```diff
@@ -371,6 +372,47 @@ func TestPersistentWorker(t *testing.T) {
...
+  isolation:
+    container:
+      image: debian:latest
+      volumes:
+        - %s:/dir-to-be-mounted
+
+unix_task:
+  show_script: ls -lah /dir-to-be-mounted
+  check_script:
+    - test -e /dir-to-be-mounted/some-file-name.txt
```
fkorotkov, line 393: Maybe it's better to test that we can mount a folder and a task can write a file to it? Seems like a more common use case to just have a persistent folder that tasks are writing to. 🤔

internal/executor/instance/container/run.go
fkorotkov: IMO we should just create it. But I'm not sure is Docker already does it. Will be great to modify the test to try to write to some non-existing mounted folder and see what will happen

edigaryev: >But I'm not sure is Docker already does it.

It doesn't, you can easily try it out like that: `docker run -it -v /non-existent:/non-existent --rm debian:latest`.

fkorotkov (approved):

PR #380, opened 2021-05-07
Use main as the default revision

fkorotkov:
All the new repos has `main` as the default branch. Since the feature is not announced and all the Cirrus Templates will be created with the new defaults it's reasonable to change it here too.


edigaryev (approved):

edigaryev:
Looks like we'll need to also rename the default branches in all https://github.com/cirrus-templates repos.

fkorotkov:
@edigaryev right! I though they are already using `main` but only the latest `graphql` was. Renamed all the branches accordingly and re-running tasks now...

PR #382, opened 2021-05-07
Use os.ModePerm when creating missing folders

fkorotkov:
Seems it's the recommended way https://stackoverflow.com/a/28448702/1390817

### DIFF
--- a/internal/executor/agent/agent.go
+++ b/internal/executor/agent/agent.go
@@ -26,7 +26,7 @@ func RetrieveBinary(
 
 	agentCacheDir := filepath.Join(cacheDir, "cirrus", "agent")
 
-	if err := os.MkdirAll(agentCacheDir, 0700); err != nil {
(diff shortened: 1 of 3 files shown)


edigaryev:
Not sure what are you trying to fix here (these bits are already [portable across systems](https://golang.org/pkg/os/#FileMode)), but I'm pretty sure we don't need to create world-writable direcrories (which `ModePerm ` does since it expands to `0o777`) for the CLI to work.

fkorotkov:
I'm testing out the isolation on Ubuntu worker here https://cirrus-ci.com/task/5412690534334464 and seeing `failed to create non-existent source directory: mkdir : no such file or directory`. Quick googling led me to that Stackoverflow post. My understanding was that `ModePerm ` is need when creating a chain of directories similar to `mkdir -p`. Will take a deeper look into the issue next week.

PR #384, opened 2021-05-12
Persistent worker with none isolation: gracefully terminate agent

### LINKED ISSUE #383: Verify that logs and processes of a canceled task are cleaned up
In case of a task running py a persistent worker without isolation. I see that in some cases `cirrus-agent-ID.log` files are left for the canceled tasks.

edigaryev:
Resolves #383.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/none/none.go
+++ b/internal/executor/instance/persistentworker/isolation/none/none.go
@@ -7,11 +7,13 @@ import (
 	"github.com/cirruslabs/cirrus-cli/internal/executor/agent"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/instance/persistentworker/pwdir"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/instance/runconfig"
+	"github.com/cirruslabs/cirrus-cli/internal/logger"
 	"github.com/otiai10/copy"
 	"os"
 	"os/exec"
 	"runtime"
 	"strconv"
+	"time"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


internal/executor/instance/persistentworker/isolation/none/none.go
```diff
@@ -74,7 +89,45 @@ func (pwi *PersistentWorkerInstance) Run(ctx context.Context, config *runconfig.
...
+		select {
+		case <-ctx.Done():
+			pwi.logger.Debugf("gracefully terminating agent with PID %d", cmd.Process.Pid)
+			_ = cmd.Process.Signal(os.Interrupt)
+		case <-runCtx.Done():
+			// agent exited
+			return
+		}
+
+		const gracefulLimit = 3 * time.Second
```
fkorotkov, line 110: Do you think it's a bit too low? Maybe give it, let's say, 30 seconds?

internal/executor/testdata/persistent-worker-graceful-termination/.cirrus.yml
```diff
@@ -0,0 +1,4 @@
+persistent_worker: {}
+
+unix_task:
+  script: sleep 86400
```
fkorotkov, line 4: What do you think of a bit more complex example? For example:

```yaml
persistent_worker: {}

unix_task:
  deattach_script: sleep 86400 &
  ps_script: ps
  sleep_script: sleep 86400
```

So in the test we can parse `pid` from `deattach` script and verify it's not longer running after the agent is killed?

edigaryev: I think we only need to check that the agent terminated gracefully here.

What you've demonstrated above belongs to the agent's tests.

internal/executor/instance/persistentworker/isolation/none/none.go
edigaryev: I believe it's already high enough since we're somewhat violating the `context.Context` expectations of an immediate termination in `Run()`.

Plus, technically it's possible to terminate uncleanly even with a 30 second timeout. Perhaps we should log these cases instead and analyze them to see if it's actually a problem?

The other option is pass a file name of the log file before it's created from the agent and ensure it's deleted from the CLI later.

fkorotkov: I'm just thinking about potential cases of executing Gradle daemon, mysql, etc. Some heavy things might take a while to get terminalted. Agreed it's not a big issue. I just was curious about the 3 seconds constant and thought it's a bit low. Will leave it up to your judgement then. 👍

fkorotkov (approved):

internal/executor/instance/persistentworker/isolation/none/none.go
edigaryev: > I'm just thinking about potential cases of executing Gradle daemon, mysql, etc. Some heavy things might take a while to get terminalted.

Note that these things will be killed (https://github.com/cirruslabs/cirrus-ci-agent/blob/061ab28d68561215957f6c92752ada0dcd55369b/internal/executor/shell.go#L65), not gracefully terminated.

Issue #377, opened 2021-05-05
Provide line number information when Starlark evaluation fails


fkorotkov:
Right now `ErrExecFailed` is reported with no information about line/column where the error happened in Starlark script.

```python
def main(ctx):
    foo = None
    foo[0]
```

Returns `exec failed: unhandled index operation NoneType[int]` which is hard to trace back to the statement where it happened.

fkorotkov:
Maybe we can also add the full stacktrace to the logs of `HookResult`.

PR #385, opened 2021-05-17
Starlark: provide backtraces in logs

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/377.


fkorotkov:
Can you also make a hook evaluation return a response rather than an RPC error here https://github.com/cirruslabs/cirrus-cli/blob/3663b18a427fc75f144600d3c4cccd453b26829a/internal/evaluator/evaluator.go#L221

fkorotkov (approved):

PR #393, opened 2021-05-20
Starlark's main(): make ctx argument optional

### LINKED ISSUE #386: Support Starlark mains withou ctx
It will be cool to see if `main` in `.cirrus.star` accpets any arguments and if so, then pass `ctx` and if `main` expects no arguments then doesn't pass `ctx` at all.

edigaryev:
Resolves #386.

### DIFF
--- a/pkg/larker/larker.go
+++ b/pkg/larker/larker.go
@@ -96,8 +96,9 @@ func (larker *Larker) Main(ctx context.Context, source string) (*MainResult, err
 			return
 		}
 
-		// Prepare a context to pass to main() as it's first argument if needed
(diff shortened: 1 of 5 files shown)


edigaryev:
@fkorotkov what do you think about making `ctx` optional for hooks too (and for `on_github_event` too in the future), to make things a little more consistent?

fkorotkov:
> @fkorotkov what do you think about making ctx optional for hooks too (and for on_github_event too in the future), to make things a little more consistent?

I'm all up for it!

fkorotkov (approved):

PR #394, opened 2021-05-20
[regression] Don't fail on missing main()

fkorotkov:


fkorotkov:
Will add a test though

pkg/larker/larker.go
```diff
@@ -117,6 +117,12 @@ func (larker *Larker) Main(ctx context.Context, source string) (*MainResult, err
 	select {
 	case mainResult = <-resCh:
 	case err := <-errCh:
+		if errors.Is(err, ErrNotFound) {
```
edigaryev, line 120: How about we keep throwing the `ErrNotFound` here and handle it in the consumers?

After all, we're calling `Main()`, which means we're expecting it to be there.

fkorotkov: It does throw. Just the result is not 'nil' but empty

edigaryev: Why?

fkorotkov: It's easier in test and other places to not put checks for `nil`

fkorotkov: Another option can be to not return `ErrNotFound` from `Main` and only return from `Hook`. If there is no main assume the list of tasks is empty.

edigaryev: > It's easier in test

It seems that it's only `TestNoMain` that fails when the returned `*MainResult` is `nil`. And frankly, this test is useless at this level of abstraction anyway, plus we already have `TestStarlarkNoMain` in `evaluator_test.go`.

How about simply removing `TestNoMain`, which will in turn allow us to continue throwing the error and foster proper error checking practices, which is a good thing?

fkorotkov: How about c7b9554 where I introduced `MainOptional` for most of the cases where absence of `main()` is totally fine.

edigaryev (approved):

Issue #397, opened 2021-05-24
Option to generate a JSON test report


fkorotkov:
Will be great to have an option to generate a machine readable test report they can be parsed and converted to Cirrus annotations

fkorotkov:
For example, `cirrus internal run --report=report.json`.

PR #417, opened 2021-07-02
Implement cirrus internal test's --report flag

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/397.

### DIFF
--- a/internal/commands/internal/test/test.go
+++ b/internal/commands/internal/test/test.go
@@ -3,6 +3,7 @@
 package test
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"github.com/cirruslabs/cirrus-cli/internal/commands/logs"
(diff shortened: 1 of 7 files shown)


internal/commands/internal/test/test.go
```diff
@@ -174,5 +254,8 @@ func NewTestCmd() *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&output, "output", "o", logs.DefaultFormat(), fmt.Sprintf("output format of logs, "+
 		"supported values: %s", strings.Join(logs.Formats(), ", ")))
 
+	cmd.PersistentFlags().StringVar(&reportFilename, "report", "",
+		"additionally write a report in Cirrus Annotation Format to this file")
```
fkorotkov, line 258: Should we add a link to the annotations repo? Or will it to weird to have a link in terminal output?

edigaryev: I've added it and it looks OK, see https://github.com/cirruslabs/cirrus-cli/pull/417/commits/79ca5ecec14f05f30a1394830bf8b088a87539d0.

Plus we can always remove it in the future when the format becomes more widely known.

fkorotkov (approved):

Issue #418, opened 2021-07-05
Unable to use when MacOS tasks are present


real-or-random:
We have MacOS tasks in our `.cirrus.yml`. These are not supported by `cirrus-cli`, which we are aware of but this is not the issue here.

The problem is that these tasks prevent us from running the Linux tasks:
```
Error: failed to create task "x86_64: macOS Catalina": failed to create instance: got nil instance which means it's probably not supported by the CLI
```

My second try was `run "x86_64: Linux (Debian stable)"` but this doesn't work because the task depends on a "Prebuild ..." task:
```
🕛 'x86_64: Linux (Debian stable)' task (ASM:no BENCH:yes BUILD:check CC:gcc CTIMETEST:yes ECDH:no ECMULTGENPRECISION:auto EXPERIMENTAL:no RECOVERY:yes SCHNORRSIG:no STATICPRECOMPUTATION:yes s
   ✅ Preparing execution environment... 2.3s
   ❌ image pull 1.0s
      Pulling image gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest...
      Failed to pull gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest: Error response from daemon: manifest for gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:l.
      
Error: Error response from daemon: manifest for gcr.io/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e:latest not found: manifest unknown: Failed to fetch "latest" from request "/v2/cirrus-ci-community/d41d8cd98f00b204e9800998ecf8427e/manifests/latest".
```

While it's possible to run `cirrus run "Prebuild ci/linux-debian.Dockerfile"`, it's unfortunately not possible to specify multiple tasks, so no luck here. The only thing that works is to edit `.cirrus.yml` and remove the MacOS tasks.

See https://github.com/bitcoin-core/secp256k1/blob/master/.cirrus.yml

Solving #31 would solve this issue. Even just skipping MacOS tasks would solve this issue.

fkorotkov:
After running `cirrus run "Prebuild ci/linux-debian.Dockerfile"` you should get the image locally built and subsequential invocation of `cirrus run --container-lazy-pull "x86_64: Linux (Debian stable)` should work theoretically. 🤔 

[...]

PR #419, opened 2021-07-05
cirrus run: keep running tasks even if there are failed ones

### LINKED ISSUE #31: Don’t fail the whole build if only one task has failed
Currently the logic in the executor's main loop is quite simple and looks like this:

https://github.com/cirruslabs/cirrus-cli/blob/3b9cb68329a9ada7f3e6c3eac0d72ab442976a4d/internal/executor/executor.go#L132-L135

It would be nice to preserve the same behavior when running tasks in the cloud.

It also seems that to do that one needs to calculate the dependencies in `Build.GetNextTask()` in a bit more complicated fashion than it currently gets away with.

edigaryev:
Resolves #31.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -30,7 +30,6 @@ linters:
   - gofmt
   - goheader
   - golint
-  - gomnd
   - gomodguard
   - goprintffuncname
   - gosec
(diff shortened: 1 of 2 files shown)


real-or-random:
Unless I'm doing something wrong in testing, this still gives me

```
Error: failed to create task "x86_64: macOS Catalina": failed to create instance: got nil instance which means it's probably not supported by the CLI
```

Maybe it's a better idea to (add a flag to?) skip macOS tasks in order to solve #418.

fkorotkov (approved):

PR #421, opened 2021-07-08
Parser: support multiple folders in cache instruction

edigaryev:
See https://github.com/cirruslabs/cirrus-ci-docs/issues/565.


pkg/parser/task/command/cache.go
```diff
@@ -38,21 +38,23 @@ func NewCacheCommand(mergedEnv map[string]string, boolevator *boolevator.Booleva
 		return nil
 	})
 
-	folderSchema := schema.String("Path of a folder to cache.")
+	folderSchema := schema.StringOrListOfStrings("Path of a folder or a list of folders to cache.")
 	cache.RequiredField(nameable.NewSimpleNameable("folder"), folderSchema, func(node *node.Node) error {
```
fkorotkov, line 42: Should it be a separate field `folders`? The question is how to make sure that either `folder` or `folders` is set. 🤔  Maybe make them optional and validate that `Fodlers` is not empty?

edigaryev: > Should it be a separate field `folders`?

We already have the `script` with the same single/multi semantics, why not continue the trend and there will be no issue with require/optional validation problems?

fkorotkov: With a script you kind of making it multiline even though it's a YAML list. IMO it's just weird to see

```yaml
cache:
  folder:
    - cache/foo
    - cache/bar
    - cache/baz
  ```

Comment (edigaryev) [reply pkg/parser/task/command/cache.go:42]: > IMO it's just weird to see

Looks similar to `script` to me:

```yaml
task:
  script:
    - make test
    - make build
```

The user doesn't know about the multiline internals anyway.

But I can split it into separate fields if you still think it should be done this way.

edigaryev: Oh, I think I've got what you mean by `script` being multiline. I'll add another field then.

edigaryev: Implemented in https://github.com/cirruslabs/cirrus-cli/pull/421/commits/575a99371b50739291b87bbcc29f1fd0528d6ae0.

pkg/parser/task/command/cache.go
```diff
@@ -39,20 +41,25 @@ func NewCacheCommand(mergedEnv map[string]string, boolevator *boolevator.Booleva
 	})
 
 	folderSchema := schema.String("Path of a folder to cache.")
-	cache.RequiredField(nameable.NewSimpleNameable("folder"), folderSchema, func(node *node.Node) error {
+	cache.OptionalField(nameable.NewSimpleNameable("folder"), folderSchema, func(node *node.Node) error {
```
fkorotkov, line 44: I think you can use the logic that was before the push with expansion of HOME and stuff. Just extract the func in a variable and reuse for both folder and folders filelds. And you can use a single folders slice variable and check if it's not empty for the validation.

edigaryev: > I think you can use the logic that was before the push with expansion of HOME and stuff.

I'm already doing this here https://github.com/cirruslabs/cirrus-cli/pull/421/files#diff-d9b28f74aa650a725937c6d5795b22f5d217930cc229655e2841c6bf11373a4eR121-R135.

>And you can use a single folders slice variable and check if it's not empty for the validation.

Not really, because you won't be able to catch cases where both `folder` and `folders` are set.

pkg/parser/task/command/cache.go
```diff
@@ -39,20 +41,25 @@ func NewCacheCommand(mergedEnv map[string]string, boolevator *boolevator.Booleva
...
 		if err != nil {
 			return err
 		}
 
-		// https://github.com/cirruslabs/cirrus-ci-agent/issues/47
-		const homePrefix = "~/"
-		folder = strings.TrimSpace(folder)
-		if strings.HasPrefix(folder, homePrefix) {
-			folder = "$HOME/" + strings.TrimPrefix(folder, homePrefix)
+		cache.folder = folder
```
fkorotkov, line 50: Let's remove `folder` and use only `folders like that:

```suggestion
		cache.folders = []string{folder}
```

fkorotkov (approved):

Issue #427, opened 2021-07-12
Add a warning when task name is overriden


fkorotkov:
Add a warning to the parser that a name is overriden in such cases:

```yaml
rspec_code_task:
  matrix:
    name: Code (shard 1)
    name: Code (shard 2)
    name: Code (shard 3)
```

Such config might confuse people on what to use in `depends_on`.

PR #429, opened 2021-07-13
Add a warning when task name is overriden

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/427.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -158,7 +158,7 @@ func (p *Parser) parseTasks(tree *node.Node) ([]task.ParseableTaskLike, error) {
 
 			if taskLike.Name() != "" && quickTaskName != "" {
 				p.registerIssuef(api.Issue_WARNING, treeItem.Line, treeItem.Column,
-					"giving a task multiple names (%q and %q) can be ambiguous",
+					"task's name %q will be overridden by %q",
 					quickTaskName, taskLike.Name())
 			}
 
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -109,6 +109,7 @@ func (p *Parser) registerIssuef(level api.Issue_Level, line int, column int, for
 	})
 }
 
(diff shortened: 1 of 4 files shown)


pkg/parser/parser.go
```diff
@@ -148,6 +149,20 @@ func (p *Parser) parseTasks(tree *node.Node) ([]task.ParseableTaskLike, error) {
...
+
+			if rn, ok := key.(*nameable.RegexNameable); ok {
+				if rn.FirstGroupOrDefault(treeItem.Name, "") != "" {
+					hasQuickTaskName = true
+				}
+			}
+
+			if taskLike.Name() != "" && hasQuickTaskName {
+				p.registerIssuef(api.Issue_WARNING, treeItem.Line, treeItem.Column,
+					"giving a task multiple names can be ambiguous")
```
fkorotkov, line 163: Let's maybe include `taskLike.Name()` in the warning message? Like `Task already has name defined as "%s". No need to use "*_task syntax"` or something similar?

edigaryev: I don't think we need this since these messages are contextual, i.e. they are displayed right on the task that caused them.

fkorotkov: I don't think in this situation it will clear for a user why there are multiple names given for a task. IMO some human friendly message will better. This way user can map that there are tow places which define name:  'name: ' and '<name>_test' thing.

edigaryev: Improved in https://github.com/cirruslabs/cirrus-cli/pull/429/commits/491304a751b19b19f2801894b19fba9eebcd9529.

edigaryev:
> Let's maybe mention that `quickTaskName` is ignored actually? It's not multiple names, it's just one is being overriden.

See https://github.com/cirruslabs/cirrus-cli/pull/429/commits/adec0bcf5f2dfe0a01f0d88cfd0291f79c032df5.

fkorotkov (approved):

PR #430, opened 2021-07-14
Fixed parsing of command for additional containers

fkorotkov:
The command should be a string which is tokenized like `CMD` in a Dockerfile.

### DIFF
--- a/pkg/parser/instance/additionalcontainer.go
+++ b/pkg/parser/instance/additionalcontainer.go
@@ -173,14 +173,15 @@ func NewAdditionalContainer(mergedEnv map[string]string, boolevator *boolevator.
 		return nil
 	})
 
-	commandSchema := schema.Script("Container CMD to override.")
+	commandSchema := schema.String("Container CMD to override.")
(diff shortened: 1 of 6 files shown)


edigaryev:
Is this field documented anywhere?

pkg/parser/instance/additionalcontainer.go
```diff
@@ -173,14 +174,18 @@ func NewAdditionalContainer(mergedEnv map[string]string, boolevator *boolevator.
...
+		command, err := node.GetExpandedStringValue(mergedEnv)
 		if err != nil {
 			return err
 		}
 
-		ac.proto.Command = command
+		// tokenize the command
+		ac.proto.Command, err = shlex.Split(command)
+		if err != nil {
+			return err
```
edigaryev, line 187: We should wrap this in a `node.ParserError()` to get a rich error.

Example here: https://github.com/cirruslabs/cirrus-cli/blob/f23edc6fa4cd1e34174b1f119e6dc5e90079ea77/pkg/parser/instance/additionalcontainer.go#L170

fkorotkov: fixed in 63ab2e0

edigaryev (approved):

PR #434, opened 2021-07-16
cirrus run: new --affected-files flag

### LINKED ISSUE #433: How can I test `changesInclude`?
Can't figure out how to trigger it locally. Seems it always fails.

edigaryev:
Resolves #433.


fkorotkov (approved):
This LGTM but what do you thing of adding `--affected-files-since-ref` flag that can pass a Git ref and then we can use `go-git` to auto-detect the changes?

PR #441, opened 2021-07-23
Add --print option to validate

abravalheri:
Hi guys, thank you very much again for the awesome tools. I am really excited with the new Starlak builds.

With this PR I propose adding adding a `--print` flag to the `validate` command of Cirrus CLI. This idea is to allow users to inspect what they are creating in Starlak (some small quality of life improvement). What do you think about this idea?

I am not very experience with Go, but I gave it a try, hopefully I didn't do anything wrong.

I did manage to run `go test -v internal/commands/validate_test.go` and check the unit test I have added, but I am still trying to find a way of compiling the project locally into a single file executable...

### DIFF
--- a/internal/commands/validate/validate.go
+++ b/internal/commands/validate/validate.go
@@ -19,6 +19,7 @@ var ErrValidate = errors.New("validate failed")
 // General flags.
 var validateFile string
 var environment []string
+var shouldPrint bool
 
 func additionalInstancesOption(stderr io.Writer) parser.Option {
 	// Try to retrieve additional instances from the Cirrus Cloud
@@ -89,6 +90,10 @@ func validate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if shouldPrint {
+		fmt.Fprint(cmd.OutOrStdout(), configuration)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/commands/validate/validate.go
```diff
@@ -104,6 +109,8 @@ func NewValidateCmd() *cobra.Command {
 		"set (-e A=B) or pass-through (-e A) an environment variable to the Starlark interpreter")
 	cmd.PersistentFlags().StringVarP(&validateFile, "file", "f", "",
 		"use file as the configuration file (the path should end with either .yml or ..star)")
+	cmd.PersistentFlags().BoolVarP(&shouldPrint, "print", "p", false,
+		"print the configuration as YAML (useful for debugging Starlak files)")
```
edigaryev, line 113: It's called Starlark:

```suggestion
		"print the configuration as YAML (useful for debugging Starlark files)")
```

Note that there are also other instances of the same typo in this changeset.

internal/commands/validate/validate.go
```diff
@@ -89,6 +90,10 @@ func validate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if shouldPrint {
```
edigaryev, line 93: Consider moving this block before the parsing is done, so that when the `cirrus validate --print` is invoked with an invalid config, the parsing errors are printed after the potentially huge configuration, so it's easier to find them in the terminal output.

edigaryev (changes requested):
LGTM overall, thanks for contributing.

We need to fix the spelling, though, and there's also a small nit about the terminal output order.

abravalheri:
Thank you very much for the review @edigaryev. I updated the PR to reflect your comments 😄

edigaryev (approved):

abravalheri:
Thank you very much guys!

PR #445, opened 2021-07-23
Allow starlark main to return a list of tuples

abravalheri:
Cirrus interpretation of the YAML file allows repeated keys.
While standard YAML (that forbids repeated keys) maps very nicely to Starlak, this feature cannot be achieved if users are constrained to return either a dict or a list of task (dict) objects.

This PR proposes allowing a list of tuples `[(key, value)]` as an equivalent to returning a dict from `main` in a Starlark script, to circumvent this limitation.

*Motivation*: until recently, Cirrus web interface was displaying warning for some scenarios where repetition was not being used (https://github.com/cirruslabs/cirrus-ci-docs/issues/887). I also don't know if some instructions can be prefixed (e.g. `docker_builder`, it is not clear on the [docs](https://cirrus-ci.org/guide/docker-builder-vm/))

Returning dicts (or equivalent) is important when the users want to have shared env vars/container definition, and other top level instructions such as `pipe`, `docker_builder`, etc. An example of repeated top level keys can be found in the very own `.cirrus.yml` file of this repository (3 `docker_builder` and 3 `task` instructions).

---
A little disclaimer: I am not an experienced Go programmer (in fact pretty much using it for the first time in the PRs to cirrus-cli), so any feedback/suggestions/guidance is very welcome.

---
Quick question: does the CirrusCI executor uses the `larker` package in this repository to parse/run the starlarker scripts, so changing it here means things automatically change in Cirrus CI, or that is a completely different codebase that would require unrelated changes?

### DIFF
--- a/internal/commands/validate_test.go
+++ b/internal/commands/validate_test.go
@@ -89,32 +89,3 @@ func TestValidatePrintFlag(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Contains(t, buf.String(), string(validConfig))
 }
-
-var repeatedKeysStarlark = `
-def task(name, script):
-		return {'name': name, 'script': script}
-
-def main():
-		return [
-			('container', {'image': 'debian:latest'}),
-			('task', {'name': 'task1', 'script': True}),
-			task('task2', True),
-			('task', {'name': 'task3', 'script': True}),
-			task('task4', True)
-		]
-`
-
-func TestValidateStarlarkWithRepeatedKeys(t *testing.T) {
-	testutil.TempChdir(t)
-
-	config := []byte(repeatedKeysStarlark)
-	if err := ioutil.WriteFile(".cirrus.star", config, 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	command := commands.NewRootCmd()
-	command.SetArgs([]string{"validate", ""})
-	err := command.Execute()
-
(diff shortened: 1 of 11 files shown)


abravalheri:
I suppose an alternative would be just concatenating the YAML representations of each element in the case `main` returns a list... but then we also need to change the output of the [`task` helper](https://github.com/cirrus-modules/helpers) to include a [root key](https://stackoverflow.com/questions/18411145/root-object-or-no-what-is-the-best-practice-for-api-responses), i.e., instead of `task(name="name", ...)` returning `{"name": "name", ...}`, it would have to return `{"task": {"name": "name", ...}}`.

edigaryev (changes requested):
Thanks for your contribution! This is probably the most straightforward way to get around the Starlark limitation in regards to duplicate dict keys.

I've highlighted a couple of issues, especially note the `validate_test.go` one: once resolved, your PR will be able to be cleanly rebased on `master` (without resorting to manually resolving any conflicts) 🙌

abravalheri:
Thank you very much for the review @edigaryev.

I have changed the implementation in `convertmain.go` to use the type assertion `GoString` and a lenght check. Hopefully this is what you had in mind...

I also moved the tests to be more specific as you have suggested.

edigaryev (approved):

PR #450, opened 2021-08-06
Improve main() return handling

edigaryev:
This is a follow-up to https://github.com/cirruslabs/cirrus-cli/pull/445 that fixes potential panic (in https://github.com/cirruslabs/cirrus-cli/commit/d905d322131cf8e110b902e28e483f2ad9597afa) and makes handling of the `list` returned from `main()` more stricter, which is almost always a good idea for parsers, unless the very specific behavior is desired (see [The Harmful Consequences of the Robustness Principle](https://datatracker.ietf.org/doc/html/draft-iab-protocol-maintenance)).

Some explanation of the commits:

* https://github.com/cirruslabs/cirrus-cli/commit/fd6cb8962af7fde91f9b9ac106c8fde4c1068801 makes more evident to the users that we only accept `tuple`'s of certain characteristics by throwing an error
* https://github.com/cirruslabs/cirrus-cli/commit/38700c1d89b719c3c2c35d7fce77530151e7ecbf makes more evident to the users that we only accept `dict`'s and `tuple`'s in `list`'s returned from `main()` by throwing an error
* https://github.com/cirruslabs/cirrus-cli/commit/218289d8714003e305e269da59b5cc19a27ef49a is a bit controversial one, because this behavior was explicitly made possible in https://github.com/cirruslabs/cirrus-cli/pull/445 (see `TestMainReturnsMixedListTuplesAndDicts ` test-case), but it seems that this mixing makes things less straightforward to the reader, plus you can always do:
  ```python
  def main():
      return [
          ('container', {'image': 'debian:latest'}),
          ('task', task('task2', True)),
      ]
  ```

  ...instead of:

  ```python
  def main():
      return [
          ('container', {'image': 'debian:latest'}),
          task('task2', True),
      ]
  ```

  @abravalheri what do you think?


abravalheri:
Hi @edigaryev, yes, I think being more restrictive is not a problem and the list of tuples seems to allow the same feature parity allowed by the YAML config.

My previous implementation was kind of relying on the second check of the generated YAML-equivalent code and in the fact that people could use the `cirrus validate --print` to figure out what was going wrong. But I completely agree that raising an error earlier is better.

Question: it is still allowed to just return a dictionary, right? (e.g. `return {'container: {'image': 'debian:latest'}, 'task': {...}}`)

edigaryev:
> Question: it is still allowed to just return a dictionary, right? (e.g. `return {'container: {'image': 'debian:latest'}, 'task': {...}}`)

Currently yes, but we're thinking about slowly deprecating it, because the `list` of `tuple`'s approach is superior to it, and showing [how to do the same thing](https://cirrus-ci.org/guide/programming-tasks/#main)  using 3 different options might be too much.

fkorotkov (approved):

PR #452, opened 2021-08-21
Parser: hash files used in the Dockerfile's ADD and COPY instructions

edigaryev:
Things that are still need to be implemented:

* provide proper YAML line/column context to the emitted errors and warnings, so that they'll be attached to the corresponding task's `dockerfile:` field

Resolves https://github.com/cirruslabs/cirrus-cli/issues/447.

PR #453, opened 2021-08-23
Parser: wrap the final file system in a caching layer

edigaryev:
Should ease GitHub API usage in https://github.com/cirruslabs/cirrus-cli/pull/452 a bit.

### DIFF
--- a/internal/commands/internal/test/test.go
+++ b/internal/commands/internal/test/test.go
@@ -1,3 +1,4 @@
+//go:build linux || darwin || windows
 // +build linux darwin windows
 
 package test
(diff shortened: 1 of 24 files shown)


edigaryev:
Also added a small, built-in 16-entry LRU cache for GitHub filesystem in https://github.com/cirruslabs/cirrus-cli/pull/453/commits/08fa78bb5627187c3a7b985fc58899442554d0df.

This should make things twice as fast in worst case, because without the cache we would do two similar GitHub API requests here:

https://github.com/cirruslabs/cirrus-cli/blob/ceebbda4d4ae9b61ad504592ecfd05f1a816fff2/pkg/parser/dockerfile.go#L169-L172

...yet a single API call returns the data needed for both `Get()` and `ReadDir()`.

fkorotkov (approved):

PR #454, opened 2021-08-24
Parser: register additional instances

edigaryev:
To avoid getting `protoregistry.NotFound` from `anypb.UnmarshalNew()`.

Also service tasks generation wasn't previously working when invoked from the Cirrus Cloud due to that, so now we disable it explicitly.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -275,7 +275,7 @@ func (p *Parser) Parse(ctx context.Context, config string) (result *Result, err 
 	}
 
 	// Create service tasks
-	if p.noServiceTasks {
+	if !p.noServiceTasks {
 		serviceTasks, err := p.createServiceTasks(protoTasks)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


pkg/parser/parser.go
```diff
@@ -257,11 +275,13 @@ func (p *Parser) Parse(ctx context.Context, config string) (result *Result, err
 	}
 
 	// Create service tasks
-	serviceTasks, err := p.createServiceTasks(protoTasks)
-	if err != nil {
-		return nil, err
+	if p.noServiceTasks {
```
fkorotkov, line 278: Should it be `!p.noServiceTasks`? 🤔

edigaryev: Fixed in https://github.com/cirruslabs/cirrus-cli/pull/454/commits/d529e4d497f08a271b6aaeddd907cb28580e86ca.

fkorotkov (approved):

PR #457, opened 2021-09-13
Don't completely fail on failing to retrieve file

### LINKED ISSUE #456: Don't completely fail on failing to retrieve file.
If you have a `Dockerfile` like this:

```Dockerfile
FROM ubuntu:latest

COPY * /tmp/
```

Then the whole parsing fails with `failed to retrieve "*": file does not exist` parsing error.

fkorotkov:
Fixes #456

### DIFF
--- a/pkg/parser/dockerfile.go
+++ b/pkg/parser/dockerfile.go
@@ -135,8 +135,7 @@ func (p *Parser) calculateDockerfileHash(
 			newHash.Write(fileContents)
 			hashedAtLeastOneSource = true
 		}); err != nil {
-			p.registerIssuef(api.Issue_WARNING, dockerfileNode.Line, dockerfileNode.Column,
-				"%v %q: %v", ErrFailedToAnalyze, dockerfilePath, err)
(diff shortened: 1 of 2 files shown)


pkg/parser/dockerfile.go
```diff
@@ -135,7 +135,10 @@ func (p *Parser) calculateDockerfileHash(
 			newHash.Write(fileContents)
 			hashedAtLeastOneSource = true
 		}); err != nil {
-			return "", dockerfileNode.ParserError(err.Error())
+			p.registerIssuef(api.Issue_WARNING, dockerfileNode.Line, dockerfileNode.Column,
+				"%v %q: %v", ErrFailedToAnalyze, dockerfilePath, err)
+
+			return hex.EncodeToString(oldHash.Sum([]byte{})), nil
```
edigaryev, line 141: No need to fail fast here: there can be other files that will hash successfully and `hashedAtLeastOneSource` will be `true`.

fkorotkov: Fixed in b691e712f8c5de75c65e45187f72dfd8972709b8

edigaryev (approved):

PR #472, opened 2021-11-06
Starlark: introduce fs.isdir()

edigaryev:
Works similarly to Python [`os.path`'s `isdir()`](https://docs.python.org/3/library/os.path.html#os.path.isdir).

See https://github.com/cirruslabs/cirrus-ci-docs/issues/938.

### DIFF
--- a/pkg/larker/builtin/fs.go
+++ b/pkg/larker/builtin/fs.go
@@ -13,6 +13,7 @@ func FS(ctx context.Context, fs fs.FileSystem) starlark.StringDict {
 		"exists":  exists(ctx, fs),
 		"read":    read(ctx, fs),
 		"readdir": readdir(ctx, fs),
+		"isdir":   isdir(ctx, fs),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/larker/builtin/fs.go
```diff
@@ -86,3 +87,25 @@ func readdir(ctx context.Context, fs fs.FileSystem) starlark.Value {
...
+func isdir(ctx context.Context, fs fs.FileSystem) starlark.Value {
+	const funcName = "isdir"
+
+	return starlark.NewBuiltin(funcName, func(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+		var path string
+		if err := starlark.UnpackPositionalArgs(funcName, args, kwargs, 1, &path); err != nil {
+			return nil, err
+		}
+
+		fileInfo, err := fs.Stat(ctx, path)
```
fkorotkov, line 100: I think we need to make `Stat` for `GitHub` FS a bit smarted in case to use like that. Right now this `Stat` call will require an API call. For https://github.com/cirruslabs/cirrus-ci-docs/issues/938 use case `readdir` will be called first which will do `ReadDir`. Right now `ReadDir` will only cache results for the folder https://github.com/cirruslabs/cirrus-cli/blob/58ec216df39f52689371b69ecc1d596b08559b04/pkg/larker/fs/github/github.go#L140

I think we can make it a bit smarted and populate the cache optimistically. PTAL #473

fkorotkov (approved):

PR #474, opened 2021-11-08
GitHub FS: fileInfos cache

edigaryev:
An alternative implementation of https://github.com/cirruslabs/cirrus-cli/pull/473 that also caches information about directories.

### DIFF
--- a/pkg/larker/fs/github/github.go
+++ b/pkg/larker/fs/github/github.go
@@ -39,7 +39,7 @@ func New(owner, repo, reference, token string) (*GitHub, error) {
 	if err != nil {
 		return nil, err
 	}
-	fileInfosCache, err := lru.New(16)
+	fileInfosCache, err := lru.New(1024)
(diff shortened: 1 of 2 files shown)


pkg/larker/fs/github/github.go
```diff
@@ -36,18 +39,32 @@ func New(owner, repo, reference, token string) (*GitHub, error) {
 	if err != nil {
 		return nil, err
 	}
+	fileInfosCache, err := lru.New(16)
```
fkorotkov, line 42: Can probably have a bigger cache just for the informations without contents. 🤔

edigaryev: Indeed, see https://github.com/cirruslabs/cirrus-cli/pull/474/commits/f343965db8ac5f10976258cf8b64fa8c7ddd611b.

fkorotkov (approved):

PR #479, opened 2021-12-16
Parser: fix find() list iteration to calculate proper Dockerfile hashes

### LINKED ISSUE #478: dockerfile hash computation doesn't handle COPY of directories nor windows path seps
Hi,

Despite https://github.com/cirruslabs/cirrus-cli/pull/452 I see containers not being rebuilt. A bit of debugging later shows that this happens at least in the case where COPY copies a directory, rather than a file.
An example of this is (in a debugging repo):  https://cirrus-ci.com/build/5938448480600064 didn't rebuild the 'linux' container, despite a copied file changing.

After changing the COPY to copy files separately, the hash computation appears to work: https://cirrus-ci.com/build/5739968847937536

Even when using individual files I see windows dockerfiles not getting rebuilt. See https://cirrus-ci.com/build/5808529746427904 - that turns out to be because I did COPY path\to\file target/. It does appear to work after converting those to forwards slashes: https://cirrus-ci.com/build/5904551994720256

Regards,

Andres

edigaryev:
Resolves #478.

### DIFF
--- a/pkg/parser/dockerfile.go
+++ b/pkg/parser/dockerfile.go
@@ -169,7 +169,7 @@ func find(ctx context.Context, fs fs.FileSystem, path string, cb func(path strin
 
 	todo.PushBack(path)
 
-	for todoEntry := todo.Front(); todoEntry != nil; todoEntry = todoEntry.Next() {
+	for todoEntry := todo.Front(); todoEntry != nil; todoEntry = todo.Front() {
 		todoPath := todo.Remove(todoEntry).(string)
 
 		namesInDir, err := fs.ReadDir(ctx, todoPath)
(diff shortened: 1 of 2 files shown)


pkg/parser/dockerfile.go
```diff
@@ -169,7 +169,7 @@ func find(ctx context.Context, fs fs.FileSystem, path string, cb func(path strin
 
 	todo.PushBack(path)
 
-	for todoEntry := todo.Front(); todoEntry != nil; todoEntry = todoEntry.Next() {
+	for todoEntry := todo.Front(); todoEntry != nil; todoEntry = todo.Front() {
 		todoPath := todo.Remove(todoEntry).(string)
```
fkorotkov, line 173: Maybe simplify the whole loop into this:

```suggestion
	for todo.Len() > 0 {
		todoPath := todo.Remove(todo.Front()).(string)
```

fkorotkov (approved):

PR #483, opened 2022-02-14
docs: PERSISTENT-WORKERS.md - add [sudo] hint to basic commands

RandyMcMillan:
Most CI sequences do an `apt-get update` command.
If the persistent worker doesn't have `sudo` rights
these type of commands - when executed - fail.
It is not obvious what the reason is initially, 
this hint will save others time trying to figure this out.

### DIFF
--- a/PERSISTENT-WORKERS.md
+++ b/PERSISTENT-WORKERS.md
@@ -13,7 +13,7 @@ Follow the instruction in the [INSTALL.md](INSTALL.md) but note that Docker or P
 The simplest invocation looks like this:
 
 ```
-cirrus worker run --token <poll registration token>
+[sudo] cirrus worker run --token <poll registration token>
 ```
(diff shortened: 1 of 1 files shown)


edigaryev:
>It is not obvious what the reason is initially

It's [principle of the least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

>this hint will save others time trying to figure this out

Your approach is nice, because it's simple, but I don't feel good about throwing `sudo` without explaining why it might be needed.

What do you think about an alternative approach to hint this in https://github.com/cirruslabs/cirrus-cli/pull/484? Could it be made better?

RandyMcMillan:
@edigaryev - i will read and give some feed back asap. :)

fkorotkov:
Closing in favor of #484

PR #491, opened 2022-03-22
Parser: fix map merging when using aliases

edigaryev:
This one focuses only on aliases, because merging everything per YAML spec is not easy as it sounds at this moment:

* we have to exclude `always`, `on_failure`, `upload_caches` and possibly others
* we have to support old behavior where duplicate key contents are merged together, e.g.: https://github.com/cirruslabs/cirrus-cli/blob/9f9dfcda39cbbd9ac260c1a30c3023a1c3111ea5/pkg/parser/testdata/via-rpc/new-deduplicator-same-type.yml#L6-L12

Resolves https://github.com/cirruslabs/cirrus-ci-docs/issues/980.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/boolevator"
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/issue"
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/modifier/matrix"
+	"github.com/cirruslabs/cirrus-cli/pkg/parser/modifier/merger"
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/nameable"
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/node"
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/parseable"
@@ -218,6 +219,9 @@ func (p *Parser) Parse(ctx context.Context, config string) (result *Result, err 
 
 	// Run modifiers on it
 	if err := matrix.ExpandMatrices(tree); err != nil {
+		return nil, err
+	}
+	if err := merger.MergeMapEntries(tree); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/parser/parser.go
```diff
@@ -220,6 +221,9 @@ func (p *Parser) Parse(ctx context.Context, config string) (result *Result, err
 	if err := matrix.ExpandMatrices(tree); err != nil {
 		return nil, err
 	}
+	if err := merger.MergeMapEntries(tree); err != nil {
+		return nil, err
+	}
```
fkorotkov, line 226: Have you investigate an option to do it while yaml to node conversion. There is already some logic about alias around this line https://github.com/cirruslabs/cirrus-cli/blob/378b674f84f6092b74c33cc22e23422c6594fdb1/pkg/parser/node/yaml.go#L126

edigaryev: I've initially started with the approach you're suggesting and found out that a separate single-pass makes things simpler to reason about.

It's not enough to implement some logic in `MergeMapsOrOverwrite`, `convert()` in `yaml.go` will need to be changed too because we'll need to re-evaluate the logic in `merger.go` every time a new map element is inserted and overwrite `Children` slice every time such case is detected.

fkorotkov (approved):

fkorotkov:
This example seems now working any more:

```yaml
.common:
  cdTools: &cdTools cd-tools:latest
  dockerBuildContainer: &dockerBuildContainer
    image: *cdTools
    cpu: 1
    memory: 1Gb
    additional_containers:
      - name: docker
        privileged: true
        cpu: 2
        memory: 10Gb
        image: docker:18.09.8-dind
        port: 2375
        env:
          DOCKER_TLS_CERTDIR: "" # Disabling TLS on daemon
  dockerTestContainer: &dockerTestContainer
    <<: *dockerBuildContainer
    additional_containers:
      - memory: 20Gb

env:
  DOCKER_HOST: tcp://localhost:2375

ResolveUnitTests_task: &ContainerizedTestTask
  container: *dockerBuildContainer
  unit_script: ./scripts/unit-tests.sh

ResolveIntegrationTests_task:
  <<: *ContainerizedTestTask
  container: *dockerTestContainer
  env:
    DOCKERFILE: integration.Dockerfile
  integration_script: ./scripts/integration-test.sh
```

You can check it in the Cloud tests in `singleGKEContainer3.yml` file.

PR #492, opened 2022-03-24
Fix SA1019 linter error (deprecated strings.Title())

edigaryev:
Also bump Go version to 1.18.


fkorotkov (approved):

edigaryev:
Seems like https://github.com/golangci/golangci-lint/issues/2649 is blocking us from switching to Go 1.18 right away, so postpone this for now.

PR #493, opened 2022-03-28
Tart isolation

edigaryev:
This allows using `Virtualization.Framework` wrapper as an alternative to Parallels when virtualizing macOS running on Apple silicon.

Things that still need to be implemented:

* `cirrus run` support (the issue is mostly about communicating with the executor's RPC service)
  * figure out how to avoid binding to `0.0.0.0`


internal/executor/instance/persistentworker/isolation/parallels/parallels.go
```diff
@@ -73,122 +67,9 @@ func (parallels *Parallels) Run(ctx context.Context, config *runconfig.RunConfig
...
-		if err := monitorCtx.Err(); err != nil {
-			return err
-		}
-
-		return fmt.Errorf("%w: failed to run agent on VM %q: %v", ErrFailed, vm.Ident(), err)
-	}
-
-	return nil
+	return remoteagent.WaitForAgent(ctx, parallels.logger, ip,
+		parallels.sshUser, parallels.sshPassword, parallels.agentOS, "amd64",
```
fkorotkov, line 71: I think it's safe to pass `runtime.GOARCH` instead of `"amd64"` here. Parallels on M1 only can ran Arm VMs.

internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -0,0 +1,89 @@
...
+
+type Tart struct {
+	logger      logger.Lightweight
+	vmName      string
+	sshUser     string
+	sshPassword string
+}
+
+func New(vmName string, sshUser string, sshPassword string, opts ...Option) (*Tart, error) {
+	parallels := &Tart{
```
fkorotkov, line 26: should it be called `tart`?

internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -0,0 +1,89 @@
...
+			continue
+		}
+
+		break
+	}
+
+	tart.logger.Debugf("IP %s retrieved from VM %s, running agent...", ip, vm.Ident())
+
+	return remoteagent.WaitForAgent(ctx, tart.logger, ip,
+		tart.sshUser, tart.sshPassword, "darwin", "arm64", config, true)
```
fkorotkov, line 80: Same here. `runtime.GOARCH` should work but it's fine to to leave it as is since tart works only with Apple Silicon

internal/executor/instance/persistentworker/isolation/tart/tart.go
edigaryev: Indeed, fixed in https://github.com/cirruslabs/cirrus-cli/pull/493/commits/f477004225562119a49ef3349d24ecfbc26353a0.

internal/executor/instance/persistentworker/isolation/parallels/parallels.go
edigaryev: >Parallels on M1 only can ran Arm VMs.

I think that's exactly the reason why it should be pegged to a specific architecture instead of being dynamic.

We explicitly signal to the reader that it only works on `amd64`.

edigaryev: Oh, I've re-read the sentence and assuming that we're going to support Parallels on M1 it makes sense to switch `runtime.GOARCH` here.

But are we going to support it? Otherwise we should instead place a guard somewhere here to prevent running on non-`amd64`: https://github.com/cirruslabs/cirrus-cli/blob/c1b4602d37df924a3177b9c8272c1424157db02f/internal/executor/instance/persistentworker/persistentworker.go#L27-L29

internal/executor/instance/persistentworker/isolation/tart/tart.go
edigaryev: Let's keep it this way as it is more explicit.

internal/executor/instance/persistentworker/isolation/parallels/parallels.go
fkorotkov: Sounds good. Let's error that it's not supported and please use tart.

fkorotkov (approved):

PR #494, opened 2022-03-31
Parser: introduce repeatable fields

edigaryev:
An alternative take on https://github.com/cirruslabs/cirrus-cli/pull/491 that has the following advantages:

* no weird `Merged` flag and no additional pre-processing (also addresses https://github.com/cirruslabs/cirrus-cli/pull/491#pullrequestreview-917299562)
* addresses https://github.com/cirruslabs/cirrus-cli/pull/491#issuecomment-1078547613 without a need for a separate white-list of YAML keys that can be repeated

And the following disadvantages:

* duplicate scripts will be silently overridden, e.g.: https://github.com/cirruslabs/cirrus-cli/blob/3adf7b852e3d0fbc72f366314c2739dfb73a0534/pkg/parser/testdata/via-rpc-invalid/validation-duplicate-commands-of-same-type.yml#L4-L6


pkg/parser/parseable/parseable.go
```diff
@@ -59,22 +60,37 @@ func (parser *DefaultParser) Parse(node *nodepkg.Node, parserKit *parserkit.Pars
...
-				seenFields[field.name.MapKey()] = struct{}{}
+		field := parser.NormalField(child.Name)
+		if field == nil {
+			continue
+		}
 
-				if err := field.onFound(child); err != nil {
-					return err
-				}
+		redefinition, hasRedefinition := redefinitions[child.Name]
```
fkorotkov, line 80: `hasRedefinition` will always be true. I think it can be omitted.

pkg/parser/parseable/default.go
```diff
@@ -43,3 +52,13 @@ func (parser *DefaultParser) Collectible() bool {
 func (parser *DefaultParser) SetCollectible(value bool) {
 	parser.collectible = value
 }
+
+func (parser *DefaultParser) NormalField(name string) *Field {
```
fkorotkov, line 56: Should it be called something like `FindFieldByName`?

edigaryev: Renamed in https://github.com/cirruslabs/cirrus-cli/pull/494/commits/72f68def2656d04a50fb66a969513fe6fc95cbdc.

fkorotkov (approved):

PR #495, opened 2022-04-02
Use SSH port forwarding to allow VM to access executor's RPC server

edigaryev:
This PR:

* adds support for `tart:` isolation for `persistent_worker:` instances
* resolves the issue raised in https://github.com/cirruslabs/cirrus-cli/pull/493#issue-1183569056 (Tart isolation requires binding to `0.0.0.0` when using `cirrus run`) by using [SSH's TCP/IP port forwarding](https://datatracker.ietf.org/doc/html/rfc4254#section-7)


internal/executor/endpoint/endpoint.go
```diff
@@ -0,0 +1,44 @@
+package endpoint
+
+type Endpoint interface {
+	Container() string
+	Direct() string
```
fkorotkov, line 5: Could you please add comments explaining the difference between the endpoints so there is no need to check out the usage in order to understand it.

internal/executor/instance/persistentworker/remoteagent/remoteagent.go
```diff
@@ -148,3 +161,39 @@ func WaitForAgent(
...
+	vmListener, err := cli.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return "", fmt.Errorf("%w: failed to set-up SSH port forwarding: %v", ErrFailed, err)
+	}
+
+	go func() {
+		for {
+			vmConn, err := vmListener.Accept()
+			if err != nil {
+				logger.Debugf("failed to accept connection from the VM on forwarded port: %v", err)
```
fkorotkov, line 176: Is there a need to wait a bit? 🤔

internal/executor/instance/persistentworker/remoteagent/remoteagent.go
```diff
@@ -148,3 +161,39 @@ func WaitForAgent(
 
 	return nil
 }
+
+func forwardViaSSH(cli *ssh.Client, logger logger.Lightweight, endpoint string) (string, error) {
+	// Expose local RPC service to the VM via SSH port forwarding
+	vmListener, err := cli.Listen("tcp", "127.0.0.1:0")
```
fkorotkov, line 167: Nice one! 🎉 Appeared to be easier than we though.

internal/executor/instance/persistentworker/remoteagent/remoteagent.go
edigaryev: > Is there a need to wait a bit? 🤔

The `Accept()` is blocking, so it should be fine.

Anyway, thanks for noticing this, I've realized we don't handle `io.EOF` here and fixed this in https://github.com/cirruslabs/cirrus-cli/pull/495/commits/3b2def425fa79e46ca1455c88c668e007ffc1e72.

internal/executor/endpoint/endpoint.go
edigaryev: Explained in https://github.com/cirruslabs/cirrus-cli/pull/495/commits/cda040c0edd2a41eb6742eb0e826478e551241ad.

Issue #489, opened 2022-03-17
Report annotations in logs when executed in GHA


fkorotkov:
It's possible to report errors/warnings/etc via [log statements in GHA](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message). Right now the CLI itself ignores any reported annotations but we can start logging them when CLI's output format set to `github`.

edigaryev:
To clarify: this is about annotations retrieved via executor's RPC.

PR #500, opened 2022-04-08
Report annotations in logs when executed in GHA

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/489.


internal/executor/rpc/rpc.go
```diff
@@ -298,3 +298,38 @@ func (r *RPC) ReportAgentSignal(ctx context.Context, req *api.ReportAgentSignalR
...
+		switch annotation.Level {
+		case api.Annotation_NOTICE:
+			mappedLevel = echelon.AnnotationLevelNotice
+		case api.Annotation_WARNING:
+			mappedLevel = echelon.AnnotationLevelWarning
+		case api.Annotation_FAILURE:
+			mappedLevel = echelon.AnnotationLevelError
+		}
+
+		r.logger.Annotation(&echelon.Annotation{
```
fkorotkov, line 324: I don't think echelon should know anything about the annotations. From it's perspective it's just a text log entry.

Maybe add the annotation notion to the `FoldableLogsRenderer` and check here if `r.logger` is an instance of it and then print the annotation?

edigaryev: Implemented direct annotation rendering in https://github.com/cirruslabs/cirrus-cli/pull/500/commits/489c9571694a4c6f7655e7d662dd7148b2a75604.

fkorotkov (approved):

PR #507, opened 2022-05-02
Switch license from MPL-2.0 to AGPL-3.0

fkorotkov:
We are planning to open source quite a few projects which will be integrated between each other including Cirrus CLI. We believe AGPL is the best license for devtools because companies are tent to modify such tools for their needs and not always share the modifications with the community.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,373 +1,661 @@
-Mozilla Public License Version 2.0
-==================================
-
-1. Definitions
---------------
-
-1.1. "Contributor"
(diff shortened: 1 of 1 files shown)


fkorotkov:
@abravalheri @ligurio please take a look and please comment with "I agree" if you are OK with re-licensing of your contributions. Please also let us know if you have any questions.

ligurio:
@fkorotkov I agree

abravalheri:
I agree.

Issue #510, opened 2022-05-06
--lazy-pull flag


fkorotkov:
Right now `cirrus run` has `--container-lazy-pull` flag which will make the CLI not try to pull a newer image if it's already presented. Since #502 we now can run tasks in Virtual Machines which can also be pulled remotely.

I think it will make sense to rename the flag into `--lazy-pull` and propagate the behaviour to Tart.

Related to https://github.com/cirruslabs/tart/issues/36

PR #517, opened 2022-05-15
cirrus run: introduce --lazy-pull flag

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/510.

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -135,7 +135,7 @@ func run(cmd *cobra.Command, args []string) error {
 
 	// Tart-related options
 	executorOpts = append(executorOpts, executor.WithTartOptions(options.TartOptions{
-		EagerPull: !(lazyPull || tartLazyPull),
+		LazyPull: lazyPull || tartLazyPull,
 	}))
 
 	// Environment
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -32,6 +32,9 @@ var affectedFilesGitRevision string
 var affectedFilesGitCachedRevision string
 var verbose bool
 
+// Common instance-related flags.
+var lazyPull bool
+
 // Container-related flags.
 var containerBackend string
 var containerLazyPull bool
@@ -40,6 +43,9 @@ var containerLazyPull bool
 // [1]: https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment
 var dockerfileImageTemplate string
 var dockerfileImagePush bool
+
+// Tart-related flags.
+var tartLazyPull bool
 
 // Flags useful for debugging.
 var debugNoCleanup bool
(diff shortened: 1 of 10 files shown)


edigaryev:
@fkorotkov need to update `tart` binary on the test machine to include https://github.com/cirruslabs/tart/pull/70 for the tests to pass.

fkorotkov:
Just cut `0.3.0` release and it should be available shortly. I think we can add something like `brew upgrade cirruslabs/cli/tart` to the scripts to always test agains the latest one.

fkorotkov:
In the meantime I've upgraded the worked so now it should work.

edigaryev:
> Just cut `0.3.0` release and it should be available shortly. I think we can add something like `brew upgrade cirruslabs/cli/tart` to the scripts to always test agains the latest one.

Seems that [the PR](https://github.com/cirruslabs/tart/pull/70) didn't make it through yet.

fkorotkov:
I thought it got merged. My bad!

internal/commands/run.go
```diff
@@ -120,13 +126,18 @@ func run(cmd *cobra.Command, args []string) error {
...
+		EagerPull: !(lazyPull || containerLazyPull),
 		NoCleanup: debugNoCleanup,
 
 		DockerfileImageTemplate: dockerfileImageTemplate,
 		DockerfileImagePush:     dockerfileImagePush,
 	}))
 
+	// Tart-related options
+	executorOpts = append(executorOpts, executor.WithTartOptions(options.TartOptions{
+		LazyPull: lazyPull || tartLazyPull,
```
fkorotkov, line 138: Should it also be called `EagerPull` for consistency&

internal/executor/instance/persistentworker/isolation/tart/vm.go
```diff
@@ -28,6 +35,20 @@ func NewVMClonedFrom(ctx context.Context, from string, cpu uint32, memory uint32
...
+		pullLogger.Infof("Pulling virtual machine %s...", from)
+
+		if _, _, err := CmdWithLogger(ctx, pullLogger, "pull", from); err != nil {
+			pullLogger.FinishWithType(echelon.FinishTypeSucceeded)
+			return nil, err
+		}
+
+		pullLogger.FinishWithType(echelon.FinishTypeSucceeded)
+	} else {
+		pullLogger.FinishWithType(echelon.FinishTypeSkipped)
```
fkorotkov, line 49: Do we need to print anything? What do you think of not even creating the scope?

internal/commands/run.go
edigaryev: It was called like that before https://github.com/cirruslabs/cirrus-cli/pull/517/commits/ef89082c9bdf10411b553db4b0cd41336e2f27a2, where I've changed the meaning so that each newly-initialized `TartOptions` struct would reflect the defaults.

internal/executor/instance/persistentworker/isolation/tart/vm.go
edigaryev: I think it provides better transparency into how the process could be, so that users might discover a new CLI option simply by looking at the skipped steps.

internal/commands/run.go
fkorotkov: Should the container option be renamed into `LazyPull` then?

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/517/commits/241da4f91fda46ba9c754397076bbf2d1226c298.

fkorotkov (approved):

Issue #504, opened 2022-04-13
Support resource management for persistent worker mode


fkorotkov:
Let's start supporting running multiple tasks in parallel for a given worker. I propose to add a mapping to the config file with the available resources. For example:

```yaml
labels:
  tart: installed

resources:
  macos-vms: 2.0
```

Then we can modify `WorkerInfo` message that we pass in `PollRequest` to include `available_resources`:

```protobuf
message WorkerInfo {
  map<string, string> labels = 1;
  map<string, double> available_resources = 2;
}
```

And respectively `PollResponse`'s `AgentAwareTask` will include new `resources` field.

Here is an example of how it will work:

1. Worker will initially poll with the `available_resources` equal to the one in the config.
2. Once `PollResponse` is returned. Will subtract `AgentAwareTask`'s `resources` from the `available_resources` for future polls.
  -  If `AgentAwareTask`'s `resources` are empty let's assume the tasks consumes all of the available resources.
3. Worker will start executing the task and will include the updated `available_resources` in the new polls.
4. Once task finishes the worker should add the resources back to `available_resources`.

On the cloud side we'll modify the `persisten_worker` side to include the resources along the labels:

```yaml
task:
  persistent_worker:
    labels:
      os: darwin
      arch: arm64
      tart: installed
    resources:
      macos-vm: 1
    isolation:
      tart:
        image: ghcr.io/cirruslabs/macos-monterey-base:latest
  script: echo "running on-premise in a VM"
```

edigaryev:
Two thoughts:

1. We can re-use [worker's labels](https://github.com/cirruslabs/cirrus-cli/blob/master/PERSISTENT-WORKERS.md#reserved-labels) and reserve a few more like `resources.tart.concurrency`, `resources.tart.cpu` and `resources.tart.memory`.

2. I don't think we have cross-platform ability to enforce resource limits for non-VM and non-Docker tasks started by worker. I'd propose to instead specialize the resource limitations to the specific types of instances to avoid confusion.

PR #522, opened 2022-06-02
Support resource management for persistent worker mode

edigaryev:
See https://github.com/cirruslabs/cirrus-cli/issues/504.

### DIFF
--- a/PERSISTENT-WORKERS.md
+++ b/PERSISTENT-WORKERS.md
@@ -41,6 +41,9 @@ name: "MacMini-Rack-1-Slot-2"
 
 labels:
   connected-device: iPhone12ProMax
+
+resources:
+  iphone-devices: 1
 ```
 
 Currently configuration files support the same set of options exposed via the command-line flags, while the rest of the options can only be configured via configuration file and are documented here.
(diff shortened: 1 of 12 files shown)


PERSISTENT-WORKERS.md
```diff
@@ -74,6 +77,43 @@ log:
...
+resources:
+  tart-vms: 2
+```
+
+...a worker won't run more than 2 tasks simultaneously from the following `.cirrus.yml`:
+
+```yaml
+persistent_worker:
+  resources:
+    tart-vms: 1
```
fkorotkov, line 102: Maybe add `isolation` as well here? To be a bit more close to real life use case.

pkg/parser/node/accessor.go
```diff
@@ -96,6 +97,30 @@ func (node *Node) GetStringMapping() (map[string]string, error) {
...
 
+func (node *Node) GetFloat64Mapping() (map[string]float64, error) {
+	result := make(map[string]float64)
+
+	if _, ok := node.Value.(*MapValue); !ok {
+		return nil, node.ParserError("expected a map")
+	}
+
+	for _, child := range node.Children {
+		stringValue, err := child.GetStringValue()
```
fkorotkov, line 108: I wonder if this value should be expanded with `env`. What do you think?

edigaryev: Enabled expansion https://github.com/cirruslabs/cirrus-cli/pull/522/commits/927638eda912b4609d590b2dce8602cc4e10edfe.

fkorotkov (approved):

PR #524, opened 2022-06-08
cirrus run: do not require a container backend to run

edigaryev:
Resolves https://github.com/cirruslabs/tart/issues/119.

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -54,11 +54,6 @@ func run(cmd *cobra.Command, args []string) error {
 	// https://github.com/spf13/cobra/issues/340#issuecomment-374617413
 	cmd.SilenceUsage = true
 
-	backend, err := containerbackend.New(containerBackend)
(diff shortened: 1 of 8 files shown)


torarnv:
Awesome, thanks!! :)

internal/commands/run.go
```diff
@@ -145,7 +140,14 @@ func run(cmd *cobra.Command, args []string) error {
 	)
 
 	// Container backend
-	executorOpts = append(executorOpts, executor.WithContainerBackend(backend))
+	if containerBackend != containerbackend.BackendAuto {
```
fkorotkov, line 143: Logic here and down in the PR seems to me a bit hard to understand. Maybe because container backends can be both strings and real types.

While I was reviewing the change I ended up looking a bit too much and tried a similar approach in #525 but I think it's a bit easier to wrap your head around. What do you think?

Issue #535, opened 2022-06-23
Allow credentials to be specified unencrypted


fkorotkov:
Once https://github.com/cirruslabs/cirrus-ci-docs/issues/979 is added let's allow to specify `gcp_credentials` like this:

```yaml
gcp_credentials:
  workload_identity_provider: projects/123456789/locations/global/workloadIdentityPools/my-pool/providers/my-provider
  service_account: 'my-service-account@my-project.iam.gserviceaccount.com'
```

This should result in credentials being populated with a JSON representation of the child node. For the example it will be:

```json
{
  "workload_identity_provider": "projects/123456789/locations/global/workloadIdentityPools/my-pool/providers/my-provider",
  "service_account": "my-service-account@my-project.iam.gserviceaccount.com"
}
```

PR #538, opened 2022-06-27
Parser: allow specifying credentials (e.g. gcp_credentials) as map

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/535.

### DIFF
--- a/pkg/parser/node/accessor.go
+++ b/pkg/parser/node/accessor.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/environment"
 	"github.com/cirruslabs/cirrus-cli/pkg/parser/boolevator"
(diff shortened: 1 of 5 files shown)


pkg/parser/node/accessor.go
```diff
@@ -43,6 +44,33 @@ func (node *Node) GetExpandedStringValue(env map[string]string) (string, error)
...
 
+func (node *Node) GetCredentials(env map[string]string) (string, error) {
+	switch node.Value.(type) {
+	case *ScalarValue:
+		return node.GetExpandedStringValue(env)
+	case *MapValue:
+		result := map[string]string{}
+
+		for _, child := range node.Children {
+			childValue, err := child.GetExpandedStringValue(env)
```
fkorotkov, line 55: I wonder if it makes sense to support nested declaration and use `GetCredentials` here as well.

One thing comes to mind here: [working with private registries](https://cirrus-ci.org/guide/linux/#working-with-private-registries):

```yaml
registry_config:
  auths:
    "container-registry.oracle.com":
      auth: ENCRYPTED[...]
```

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/538/commits/4fa0df370b3bbfd61a4200d2e7118f5e67230166.

fkorotkov (approved):

PR #540, opened 2022-06-27
Tart instance: pass --with-softnet flag when Softnet is enabled

edigaryev:
See https://github.com/cirruslabs/tart/pull/48.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/vm.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/vm.go
@@ -12,6 +12,8 @@ import (
 type VM struct {
 	ident string
 
+	softnet bool
+
 	subCtx       context.Context
 	subCtxCancel context.CancelFunc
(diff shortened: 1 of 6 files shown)


internal/executor/instance/persistentworker/isolation/tart/vm.go
```diff
@@ -24,12 +26,16 @@ func NewVMClonedFrom(
...
 	logger *echelon.Logger,
 ) (*VM, error) {
 	subCtx, subCtxCancel := context.WithCancel(ctx)
 
 	vm := &VM{
-		ident:        "cirrus-cli-" + uuid.New().String(),
+		ident: "cirrus-cli-" + uuid.New().String(),
+
+		softnet: softnet,
+
```
fkorotkov, line 38: seems `go fmt` is needed

fkorotkov (approved):

PR #545, opened 2022-07-27
Dockerfile as a CI environment support for container isolation

### LINKED ISSUE #541: Persistent workers: support Dockerfile
Hi,

I just wanted to check if there is a reason that persistent workers don't support a Dockerfile when using containers, but only specific images.

For our use-case, being able to use a Dockerfile would be great - we want to basically move some CPU intensive tests from cirrus machines to our own infrastructure - while keeping the way the tests are run as compatible as possible.

edigaryev:
Resolves #541.


fkorotkov:
It seems with the approach of using `PrebuiltImageInstance` we can only support `amd64` in the cloud since we don't yet support Arm. [Initially I implied](https://github.com/cirruslabs/cirrus-cli/issues/541#issuecomment-1169967467) that the caching will be scoped within worker and isolation. I wonder if it's a more resilient approach given the need to support Arm too, simplify control of the image, etc. 

On the other hand it will be hard to debug if something happens while building the image. Plus `PrebuiltImageInstance` looks very elegant. Let's try to brainstorm it offline and see what we can come up with.

fkorotkov:
Overall LGTM but let's make sure #551 is taking care of here. I think `CIRRUS_ARCH` is not set automatically by persistent worker instances.

fkorotkov (approved):

fkorotkov (approved):

PR #557, opened 2022-09-05
cirrus run: support --dirty for Tart-based instances

### LINKED ISSUE #554: [tart] use folder sharing for local runs
Plus it will allow to support --dirty mode.

Not that for the persistent worker isolation we can continue using "scp" because we just copy over the agent binary which is one file and not so large.

edigaryev:
Resolves #554.

Related: https://github.com/cirruslabs/tart/issues/224.


fkorotkov (approved):
As discussed offline let's merge folder sharing for dirty mode. But because folder sharing needs Ventura on both host and guest let's wait a bit because of backward compatibility.

PR #560, opened 2022-09-09
Improve command for Go installation

AlexWayfer:
1. I don't know why there are parentheses, they look redundant.
2. Add `cd -` at the end to return to the original directory (after `cd` to home at the begin).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -65,7 +65,7 @@ curl -L -o cirrus https://github.com/cirruslabs/cirrus-cli/releases/latest/downl
 If you have [Go](https://golang.org/) 1.17 or newer installed, you can run:
 
 ```
-cd && GO111MODULE=on go install github.com/cirruslabs/cirrus-cli/...@latest && cd -
+go install github.com/cirruslabs/cirrus-cli/...@latest
 ```
 
 This will build and place the `cirrus` binary in `$GOPATH/bin`.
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -65,7 +65,7 @@ curl -L -o cirrus https://github.com/cirruslabs/cirrus-cli/releases/latest/downl
 If you have [Go](https://golang.org/) 1.17 or newer installed, you can run:
 
 ```
-(cd && GO111MODULE=on go install github.com/cirruslabs/cirrus-cli/...@latest)
+cd && GO111MODULE=on go install github.com/cirruslabs/cirrus-cli/...@latest && cd -
```
edigaryev, line 68:
```suggestion
go install github.com/cirruslabs/cirrus-cli/...@latest
```

Let's cut it to just `go install github.com/cirruslabs/cirrus-cli/...@latest` since we're now using `go install` and [`GO111MODULE` is `on` by default](https://maelvls.dev/go111module-everywhere/#go111module-with-go-116) on newer Go versions.

edigaryev:
> 1. I don't know why there are parentheses, they look redundant.

The parentheses start the command in a sub-shell, which avoids changing the default directory of the invoking shell.

edigaryev (approved):

fkorotkov (approved):

INSTALL.md
AlexWayfer: Great. I didn't know this.

AlexWayfer:
> > 1. I don't know why there are parentheses, they look redundant.
> 
> The parentheses start the command in a sub-shell, which avoids changing the default directory of the invoking shell.

It doesn't work in such way in Fish Shell at least… parentheses have priority function (like in math), as I know. So, something more explicit like `sh -c` (or `fork`?) could be better. But I'm glad if it's not necessary anymore.

By the way, sorry for late response.

Issue #74, opened 2020-09-09
Failed to create a volume from working directory


AlexWayfer:
```
> cirrus run -v rubocop
gRPC server is listening at http://172.17.0.1:36311
running task rubocop (3)
Started 'rubocop'
Started 'Preparing execution environment...'
Preparing volume to work with...
Failed to create a volume from working directory: working volume creation failed: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/images/create?fromImage=gcr.io%2Fcirrus-ci-community%2Fcirrus-ci-agent&tag=v1.8.0": dial unix /var/run/docker.sock: connect: permission denied
'Preparing execution environment...' failed in 0.0s!
Error: working volume creation failed: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/images/create?fromImage=gcr.io%2Fcirrus-ci-community%2Fcirrus-ci-agent&tag=v1.8.0": dial unix /var/run/docker.sock: connect: permission denied
2020/09/10 00:29:03 working volume creation failed: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/images/create?fromImage=gcr.io%2Fcirrus-ci-community%2Fcirrus-ci-agent&tag=v1.8.0": dial unix /var/run/docker.sock: connect: permission denied
```

Repository: https://github.com/AlexWayfer/alexwayfer.name

Environment: ArchLinux. There is no `docker` permissions for regular user, we should use `sudo`. So, when and how I need to use `sudo` more correctly?

fkorotkov:
If you need to use `sudo` to run Docker could you please try to run `sudo cirrus run -v rubocop`?

AlexWayfer:
> If you need to use `sudo` to run Docker could you please try to run `sudo cirrus run -v rubocop`?

I think I can, but… is it secure?

AlexWayfer:
BTW, it works.

So, or we can find out a more secure (and better) way to do `sudo` for `cirrus-cli`, either we should add this note somewhere to README or anything else (at least for Arch Linux users).

edigaryev:
> There is no `docker` permissions for regular user, we should use `sudo`.

[...]

PR #564, opened 2022-09-29
Add a note about `sudo` for cache

AlexWayfer:
It seems not too obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,10 @@ cirrus validate
 ## Caching
 
 By default, Cirrus CLI stores blob artifacts produced by the [cache instruction](https://cirrus-ci.org/guide/writing-tasks/#cache-instruction)
-in the [user-specific cached data folder](https://golang.org/pkg/os/#UserCacheDir). [Similar to Cirrus Cloud](https://cirrus-ci.org/guide/writing-tasks/#http-cache)
-the CLI can use a caching HTTP server for more efficient sharing of cached artifacts between tasks executed on different physical hosts.
+in the [user-specific cached data folder](https://golang.org/pkg/os/#UserCacheDir).
+Please note, that if you run `cirrus` under `sudo` (for example, in ArchLinux by default), check `/root` directories.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,8 +99,10 @@ cirrus validate
 ## Caching
 
 By default, Cirrus CLI stores blob artifacts produced by the [cache instruction](https://cirrus-ci.org/guide/writing-tasks/#cache-instruction)
-in the [user-specific cached data folder](https://golang.org/pkg/os/#UserCacheDir). [Similar to Cirrus Cloud](https://cirrus-ci.org/guide/writing-tasks/#http-cache)
-the CLI can use a caching HTTP server for more efficient sharing of cached artifacts between tasks executed on different physical hosts.
+in the [user-specific cached data folder](https://golang.org/pkg/os/#UserCacheDir).
+Please note, that if you run `cirrus` under `sudo` (for example, in ArchLinux by default), check `/root` directories.
```
edigaryev, line 103: >if you run `cirrus` under `sudo` (for example, in ArchLinux by default)

Care to elaborate on why is this a default for Arch Linux?

AlexWayfer: I guess, security reasons. From the ArchLinux Wiki:

> If you want to be able to run the `docker` CLI command as a non-root user, add your user to the `docker` [user group](https://wiki.archlinux.org/title/User_group), re-login, and restart `docker.service`.

>
> Warning: Anyone added to the `docker` group is root equivalent because they can use the `docker run --privileged` command to start containers with root privileges. For more information see [[3]](https://github.com/docker/docker/issues/9976) and [[4]](https://docs.docker.com/engine/security/security/).

edigaryev: Sorry, but I fail to see the how your quote implies that the `sudo` is default/should be used on Arch Linux. As [we've previously discussed](https://github.com/cirruslabs/cirrus-cli/issues/74#issuecomment-690300414), consider running Cirrus CLI as a separate user and adding that user to the `docker` group instead if you're concerned about the `docker` attack vector.

Also, by default, the `sudo` invocation doesn't seem to actually change the `$HOME`:

```
% echo $HOME
/Users/edi
% sudo sh -c 'echo $HOME'   
/Users/edi
```

Do you have a different behavior on your system?

edigaryev:
Closing due to inactivity.

Issue #561, opened 2022-09-09
Support persistent worker registration to several endpoints


fkorotkov:
Let's support registering of the worker to several RPC targets / registration tokens. Will be useful so one worker can be registered into several organizations. When the worker will poll, it will first shuffle the possible targets / tokens and then poll each one of them until it gets a task assigned.

PR #565, opened 2022-10-03
Persistent Workers: multi-upstream support

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/561.

### DIFF
--- a/internal/commands/worker/config.go
+++ b/internal/commands/worker/config.go
@@ -1,165 +1,136 @@
 package worker
 
 import (
+	"errors"
 	"fmt"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/endpoint"
 	"github.com/cirruslabs/cirrus-cli/internal/worker"
+	"github.com/cirruslabs/cirrus-cli/internal/worker/upstream"
 	"github.com/dustin/go-humanize"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 	"gopkg.in/natefinch/lumberjack.v2"
+	"gopkg.in/yaml.v3"
 	"io"
 	"os"
-	"strconv"
 	"strings"
 )
 
-var (
-	configPath string
-	name       string
-	token      string
-	labels     map[string]string
+var ErrConfiguration = errors.New("configuration error")
 
-	// RPC-related variables.
-	rpcEndpointAddress string
+type Config struct {
+	Name  string `yaml:"name"`
+	Token string `yaml:"token"`
 
-	// Logging-related variables.
-	logLevel        string
-	logFile         string
-	logRotateSize   string
-	logMaxRotations uint
+	Labels    map[string]string  `yaml:"labels"`
+	Resources map[string]float64 `yaml:"resources"`
 
-	// Resources-related variables.
-	resources map[string]string
-)
+	Log ConfigLog `yaml:"log"`
 
-func loggingLevelsExplainer() string {
-	var levels []string
+	RPC ConfigRPC `yaml:"rpc"`
 
-	for _, level := range logrus.AllLevels {
-		levels = append(levels, `"`+level.String()+`"`)
-	}
-
-	if len(levels) == 0 {
-		panic("logging library reports no logging levels to use, this shouldn't normally happen; " +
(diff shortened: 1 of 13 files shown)


internal/worker/worker.go
```diff
@@ -171,177 +147,123 @@ func (worker *Worker) Run(ctx context.Context) error {
...
 	worker.registerTaskCompletions()
 
-	worker.logger.Debugf("polling %s", worker.rpcEndpoint)
-
 	request := &api.PollRequest{
-		WorkerInfo:   worker.info(),
-		RunningTasks: worker.runningTasks(),
+		WorkerInfo:     worker.info(upstream.WorkerName()),
+		RunningTasks:   worker.runningTasks(),
+		ResourcesInUse: worker.resourcesInUse(),
```
fkorotkov, line 184: I think the issue here is that tasks from all upstreams will be reported to every single upstream. I think initially we talked about making it a bit smarter and reporting only tasks polled from that particular upstream but resources in use from all the tasks of all upstreams.

I wonder if that make sense to try extend the newly created upstream logic or try to implement as we brainstormed before via leaving `Worker` as it was and introducing a new abstraction like `WorkerGroup` where the [original for-loop polling logic](https://github.com/cirruslabs/cirrus-cli/blob/f4ef224a7d730408cf8acb3170da230c39aac2ac/internal/worker/worker.go#L181-L204) will live. The logic in the worker group can look like this:

```golang
for {
  for _, worker := range group.workers {
    if worker.sessionToken == "" {
      if err := worker.register(subCtx); err != nil {
        worker.logger.Errorf("failed to register worker: %v", err)
      }
    }

    err := worker.poll(subCtx, group.resourcesInUse())
    if errors.Is(err, ErrShutdown) {
      return nil
    }
  }
   // the rest of waiting before next poll
}
```

This way the existing logic will be almost untouched. 🤔

internal/commands/worker/config.go
```diff
@@ -168,6 +139,42 @@ func buildWorker(cmd *cobra.Command) (*worker.Worker, error) {
...
+				ErrConfiguration)
+		}
+
+		if config.RPC.Endpoint != upstream.DefaultRPCEndpoint {
+			return nil, fmt.Errorf("%w: \"rpc:\" and \"endpoints:\" are mutually exclusive",
+				ErrConfiguration)
+		}
+	}
+
+	for _, configUpstream := range config.Upstreams {
```
fkorotkov, line 160: Probably can also add a validation that only resource aware workers can have more than 1 upstream.

internal/worker/worker.go
edigaryev: > I think the issue here is that tasks from all upstreams will be reported to every single upstream. I think initially we talked about making it a bit smarter and reporting only tasks polled from that particular upstream but resources in use from all the tasks of all upstreams.

Good catch, fixed this in https://github.com/cirruslabs/cirrus-cli/pull/565/commits/0ebeb27313b1739911cd3bb5442c5542a301c19b.

> I wonder if that make sense to try extend the newly created upstream logic or try to implement as we brainstormed before via leaving `Worker` as it was and introducing a new abstraction like `WorkerGroup` where the [original for-loop polling logic](https://github.com/cirruslabs/cirrus-cli/blob/f4ef224a7d730408cf8acb3170da230c39aac2ac/internal/worker/worker.go#L181-L204) will live. The logic in the worker group can look like this:

The current logic looks almost like the example you've showed above, with the exception of the registration step hidden in `worker.pollSingleUpstream()` (otherwise the `WorkerGroup` is essentially the same thing as `Upstream`, just differently named):

```go
for {
  for _, upstream := range worker.upstreams {
    if err := worker.pollSingleUpstream(subCtx, upstream); err != nil {
      if errors.Is(err, ErrShutdown) {
        return nil
      }

      worker.logger.Errorf("failed to poll upstream %s: %v", upstream.Name(), err)
    }
  }

  // the rest of waiting before next poll
}
```

I can move the `Register()` back there, but it further complicates the already nested loop (even the linter complains).

Also, are you sure we need to `group.resourcesInUse()`? Because this will hide the overall resources from other upstreams and they'll mistakenly schedule tasks to the worker that won't be scheduled if we show the full view of the worker.

fkorotkov (approved):

PR #567, opened 2022-10-07
When task is stopped delete it right away

fkorotkov:
There is no need to wait for the next completion handler call to clean this up

### DIFF
--- a/internal/worker/task.go
+++ b/internal/worker/task.go
@@ -115,6 +115,7 @@ func (worker *Worker) runTask(
 func (worker *Worker) stopTask(taskID int64) {
 	if task, ok := worker.tasks[taskID]; ok {
 		task.cancel()
+		delete(worker.tasks, taskID)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/worker/task.go
```diff
@@ -115,6 +115,7 @@ func (worker *Worker) runTask(
 func (worker *Worker) stopTask(taskID int64) {
 	if task, ok := worker.tasks[taskID]; ok {
 		task.cancel()
+		delete(worker.tasks, taskID)
```
edigaryev, line 118: Which problem are you trying to solve with this?

I don't think we can currently just remove the task without waiting for it to terminate, because that would be a spurious task completion in logs:

https://github.com/cirruslabs/cirrus-cli/blob/0ebeb27313b1739911cd3bb5442c5542a301c19b/internal/worker/task.go#L151-L157

Plus, before we receive the task completion the task technically is still running.

PR #570, opened 2022-10-14
Support resources via flags

fkorotkov:


internal/commands/worker/config.go
```diff
@@ -85,6 +90,15 @@ func buildWorker(cmd *cobra.Command) (*worker.Worker, error) {
 		},
 	}
 
+	config.Resources = map[string]float64{}
```
edigaryev, line 93: Do you think instantiating this in `Config{...}` above would be better?

fkorotkov: 👌 see [305fbb0](https://github.com/cirruslabs/cirrus-cli/pull/570/commits/305fbb0d21d99245eb56351bcce2e2b2e5200645)

edigaryev (approved):

Issue #571, opened 2022-10-18
Check if need to cleanup on workers startup


fkorotkov:
Sometimes a worker can exit prematurely. For example if OS crashes and computer restarts. In such cases there might be some files left. For example, local tart VMs if any were running upon crash. Let's make the worker to check for things to clean up upon startup.

PR #572, opened 2022-10-19
Use unique task id in temporary VM name

fkorotkov:
So it's easier to debug things like #571

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/tart.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/tart.go
@@ -60,7 +60,8 @@ func New(vmName string, sshUser string, sshPassword string, cpu uint32, memory u
 }
 
 func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err error) {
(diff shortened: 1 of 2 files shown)


internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -60,7 +60,8 @@ func New(vmName string, sshUser string, sshPassword string, cpu uint32, memory u
 }
 
 func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err error) {
-	vm, err := NewVMClonedFrom(ctx, tart.vmName, tart.cpu, tart.memory, config.TartOptions.LazyPull, config.Logger())
+	tmpVmName := fmt.Sprintf("cirrus-cli-%d", config.TaskID)
```
edigaryev, line 63: Do you think it's worth to keep the UUID at the end? Just in case the task ID uniqueness assumption holds false in tests or somewhere else.

fkorotkov: Don't think so but it's fine by me. See [81af546](https://github.com/cirruslabs/cirrus-cli/pull/572/commits/81af546c071717e402ee3f039e8514a23219108e)

edigaryev (approved):

PR #582, opened 2022-11-09
Executor: make sure we close the task's instance at the end of the run

edigaryev:


fkorotkov (commented):
How do you find it? When it popped up?

edigaryev:
> How do you find it? When it popped up?

It never popped up, I was going through a list of old TODOs 😅

fkorotkov (approved):

PR #583, opened 2022-11-17
Add tart-dirs parameters to support mounting custom directories

gsakun:


edigaryev:
Hello, thanks for your contribution!

Could you please elaborate a bit more on your specific use-case?

Currently `cirrus run` has a `--dirty` flag, which already utilizes the Tart's directory mounting by attaching the project directory in read-write mode.

Mounting additional directories is not compatible with Cirrus Cloud, so it would be nice to understand the need for this change before increasing the feature imparity between local & cloud runtimes.

gsakun:
> Hello, thanks for your contribution!
> 
> Could you please elaborate a bit more on your specific use-case?
> 
> Currently `cirrus run` has a `--dirty` flag, which already utilizes the Tart's directory mounting by attaching the project directory in read-write mode.
> 
> Mounting additional directories is not compatible with Cirrus Cloud, so it would be nice to understand the need for this change before increasing the feature imparity between local & cloud runtimes.

Ok,I use tart for IOS build tasks.If the --dirty parameter is used, only the current directory can be mounted.But in my usage scenario, I need to mount multiple directories.for example Mount both the dependent directory and the project directory
```
cirrus run --tart-dirs="repos:/Users/admin/.cocoapods/repos" --tart-dirs="workspace:/Users/admin/workspace ventura-xcode:14.1-RC
```

edigaryev:
@gsakun would `--tart-run-args="--dir repos:/Users/admin/.cocoapods/repos"` instead work for you?

gsakun:
> @gsakun would `--tart-run-args="--dir repos:/Users/admin/.cocoapods/repos"` instead work for you?

That would be ok.

PR #591, opened 2022-12-15
Use macOS Ventura for tests

edigaryev:
To minimize disk usage on Scaleway's M1 machine with small disk.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -47,7 +47,7 @@ task:
     labels:
       name: scaleway-m1
   env:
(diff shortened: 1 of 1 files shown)


fkorotkov:
Why is it minimizing the disk usage? The images are similar in size and `latest` tag will make updated images to get pulled.

edigaryev:
> Why is it minimizing the disk usage? The images are similar in size and `latest` tag will make updated images to get pulled.

Because no repository except for `cirrus-cli` is using such an old image. After this PR we could just delete it from the M1 machine, and thus free up the space.

fkorotkov (approved):

PR #593, opened 2022-12-16
Fixed waiting logic in pause command

fkorotkov:
We should wait for all the upstreams to report empty lists of tasks

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -232,17 +232,17 @@ func (worker *Worker) Pause(ctx context.Context, wait bool) error {
 		for {
 			response, err := upstream.QueryRunningTasks(ctx, &api.QueryRunningTasksRequest{})
 			if err != nil {
-				worker.logger.Errorf("ignoring an API error from upstream %s while waiting: %w",
-					upstream.Name(), err)
+				return fmt.Errorf("upstream %s failed while waiting: %w", upstream.Name(), err)
 			} else if len(response.RunningTasks) == 0 {
-				return nil
+				// done waiting for the current upstream
+				break
 			}
 
 			select {
 			case <-ctx.Done():
 				return nil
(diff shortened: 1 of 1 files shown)


edigaryev:
Perhaps we can get away by just adding a couple of `continue`'s in the old code:

```patch
diff --git a/internal/worker/worker.go b/internal/worker/worker.go
index 55bd237..4323e3b 100644
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -234,8 +234,9 @@ func (worker *Worker) Pause(ctx context.Context, wait bool) error {
                        if err != nil {
                                worker.logger.Errorf("ignoring an API error from upstream %s while waiting: %w",
                                        upstream.Name(), err)
+                               continue
                        } else if len(response.RunningTasks) == 0 {
-                               return nil
+                               continue
                        }
 
                        select {
```

What do you think?

fkorotkov:
I think in your case we'll need to label the outer loop to do a proper continue. IMO this code seems easier to read.

edigaryev:
> I think in your case we'll need to label the outer loop to do a proper continue. IMO this code seems easier to read.

You're right, it should be `break` instead of `continue`.

In terms of complexity, both variants contain a nested loop, but IMO the one I proposed is simpler anyway because it uses no additional variables and has more straightforward structure in terms of following the upstream iteration process. Note that in this regard, your variant will send unnecessary requests to the already interrogated upstreams. Are you OK with that?

fkorotkov:
For me personally it was hard to read but I don't have a strong opinion on this. PTAL [eeaa22c](https://github.com/cirruslabs/cirrus-cli/pull/593/commits/eeaa22c7e4d1c14cd3a1e55d4461cef3c8adfb49).

edigaryev (approved):

PR #594, opened 2022-12-20
Fix graceful termination of Tart VMs

edigaryev:
Make sure we run `tart stop` before the global context cancellation terminates the `tart run` process.

Should resolve most of the Sentry issues.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/vm.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/vm.go
@@ -14,10 +14,10 @@ type VM struct {
 
 	env map[string]string
 
-	subCtx       context.Context
-	subCtxCancel context.CancelFunc
-	wg           sync.WaitGroup
-	errChan      chan error
+	runningVMCtx       context.Context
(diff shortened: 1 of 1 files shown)


internal/executor/instance/persistentworker/isolation/tart/vm.go
```diff
@@ -36,14 +36,14 @@ func NewVMClonedFrom(
 	env map[string]string,
 	logger *echelon.Logger,
 ) (*VM, error) {
-	subCtx, subCtxCancel := context.WithCancel(ctx)
+	runningVMCtx, runningVMCtxCancel := context.WithCancel(context.Background())
```
fkorotkov, line 39: Is there a need to pass `ctx` now?

edigaryev: Yes, it is used multiple times just a couple of lines below.

fkorotkov: Should we use this context there though?

Issue #605, opened 2023-02-20
Option to use a temp mounted directory for Tart isolation


fkorotkov:
Since IO performance for mounted directories is 20-30% better than the Tart VM disk. Let's make an option for Tart isolation of a persistent worker to use a mounted directory as `CIRRUS_WORKING_DIRECTORY`. Persistent worker can create a temp folder, mount it and use for the build.

PR #608, opened 2023-03-13
Option to use a temp mounted directory for Tart isolation

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/605.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/tart.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/tart.go
@@ -38,6 +38,8 @@ type Tart struct {
 	cpu         uint32
 	memory      uint32
 	softnet     bool
+
+	mountTemporaryWorkingDirectoryFromHost bool
 }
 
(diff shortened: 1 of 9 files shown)


internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -77,6 +79,20 @@ func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err err
 	defer vm.Close()
 
 	// Start the VM (asynchronously)
+	var preCreatedWorkingDir string
+
+	if tart.mountTemporaryWorkingDirectoryFromHost {
+		tmpDir, err := os.MkdirTemp("", "")
```
fkorotkov, line 85: Do we need to clean this thing up when instance finishes?

edigaryev: Good catch, see https://github.com/cirruslabs/cirrus-cli/pull/608/commits/9e90f99b263b5494fc71c8e962443c812d2877f3.

fkorotkov (approved):

PR #625, opened 2023-05-29
Capture to Sentry when failed to run a task

fkorotkov:
To help debug Tart networking issues

### DIFF
--- a/internal/worker/task.go
+++ b/internal/worker/task.go
@@ -86,7 +86,7 @@ func (worker *Worker) runTask(
 			cirrusSentryTags["cirrus.upstream_hostname"] = upstream.Name()
 		}
 
-		cirrusSentryTagsFormatted := make([]string, len(cirrusSentryTags))
+		cirrusSentryTagsFormatted := make([]string, 0)
 		for k, v := range cirrusSentryTags {
 			cirrusSentryTagsFormatted = append(cirrusSentryTagsFormatted, fmt.Sprintf("%s=%s", k, v))
 		}
--- a/internal/worker/task.go
+++ b/internal/worker/task.go
@@ -129,7 +129,7 @@ func (worker *Worker) runTask(
 				sentry.WithScope(func(scope *sentry.Scope) {
 					scope.SetTags(cirrusSentryTags)
 					scope.SetLevel(sentry.LevelFatal)
-					sentry.CaptureException(err)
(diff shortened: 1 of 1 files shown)


internal/worker/task.go
```diff
@@ -72,16 +74,21 @@ func (worker *Worker) runTask(
...
 		// Provide tags for Sentry: upstream hostname
 		if url, err := url.Parse(upstream.Name()); err == nil {
-			cirrusSentryTags = append(cirrusSentryTags, fmt.Sprintf("cirrus.upstream_hostname=%s", url.Host))
+			cirrusSentryTags["cirrus.upstream_hostname"] = url.Host
 		} else {
-			cirrusSentryTags = append(cirrusSentryTags, fmt.Sprintf("cirrus.upstream_hostname=%s", upstream.Name()))
+			cirrusSentryTags["cirrus.upstream_hostname"] = upstream.Name()
+		}
+
+		cirrusSentryTagsFormatted := make([]string, 0)
```
edigaryev, line 89: Simpler syntax:

```suggestion
		var cirrusSentryTagsFormatted []string
```

internal/worker/task.go
```diff
@@ -72,16 +74,21 @@ func (worker *Worker) runTask(
 		}
 
 		// Provide tags for Sentry: task ID and upstream worker name
-		cirrusSentryTags := []string{
-			fmt.Sprintf("cirrus.task_id=%d", agentAwareTask.TaskId),
-			fmt.Sprintf("cirrus.upstream_worker_name=%s", upstream.WorkerName()),
+		cirrusSentryTags := map[string]string{
+			"cirrus.task_id":              strconv.FormatInt(agentAwareTask.TaskId, 10),
+			"cirrus.upstream_worker_name": upstream.WorkerName(),
 		}
```
edigaryev, line 80: What do you think about calling:

```go
sentry.ConfigureScope(func(scope *sentry.Scope) {
	scope.SetTags(cirrusSentryTags)
})
```

...here, to avoid repeating `scope.SetTags(cirrusSentryTags)` multiple times below?

fkorotkov: We'll need to not forget pop the scope. That's why I opted into just doing everything at once.

fkorotkov: Plus I think it might conflict when we run multiple tasks in parallel

edigaryev (approved):

PR #629, opened 2023-06-08
EvaluateRequest: support new file system abstraction

edigaryev:
Needed for https://github.com/cirruslabs/cirrus-ci-docs/issues/791.

### DIFF
--- a/internal/evaluator/evaluator.go
+++ b/internal/evaluator/evaluator.go
@@ -90,17 +90,7 @@ func (r *ConfigurationEvaluatorServiceServer) EvaluateConfig(
 		yamlConfigs = append(yamlConfigs, request.YamlConfig)
 	}
 
-	var fs fs.FileSystem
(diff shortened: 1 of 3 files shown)


internal/evaluator/evaluator.go
```diff
@@ -306,3 +265,20 @@ func TransformAdditionalInstances(
 
 	return additionalInstances, nil
 }
+
+func convertFS(apiFS *api.FileSystem) (fs fs.FileSystem, err error) {
+	fs = failing.New(ErrNoFS)
```
fkorotkov, line 270: Should it be populated with `request.FilesContents` for backward compatibility?

edigaryev: It seems that we only use it tests, so turns out no backwards compatibility is needed.

fkorotkov (approved):

PR #634, opened 2023-07-19
Tart executor: provide breadcrumbs

edigaryev:
Hopefully this will allow us to diagnose the root cause of PERSISTENT-WORKERS-12.

See also https://github.com/cirruslabs/cirrus-ci-docs/issues/1160#issuecomment-1639727838.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/tart.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/tart.go
@@ -8,6 +8,7 @@ import (
 	"github.com/cirruslabs/cirrus-cli/internal/executor/instance/runconfig"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/platform"
 	"github.com/cirruslabs/cirrus-cli/internal/logger"
+	"github.com/getsentry/sentry-go"
 	"github.com/google/uuid"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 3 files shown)


internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -151,9 +152,20 @@ func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err err
...
-	return remoteagent.WaitForAgent(ctx, tart.logger, ip,
+	addTartListBreadcrumb()
+	addDHCPDLeasesBreadcrumb()
+
+	err = remoteagent.WaitForAgent(ctx, tart.logger, ip,
 		tart.sshUser, tart.sshPassword, "darwin", "arm64",
 		config, true, hooks, preCreatedWorkingDir)
+	if err != nil {
+		addTartListBreadcrumb()
+		addDHCPDLeasesBreadcrumb()
```
fkorotkov, line 163: I think this will permanently add the breadcrumbs. Let's just report a single event here with the breadcrumbs?

edigaryev: >I think this will permanently add the breadcrumbs.

What do you mean by "permanently"? The breadcrumbs will only be included if an exception occurs and will be attached to that exception.

fkorotkov: The persistent worker process is running constantly and executes different tasks. In this situation the next tasks will have breadcrumbs of the previous task.

edigaryev: Ohh, good catch! Fixed in https://github.com/cirruslabs/cirrus-cli/pull/634/commits/82549d82c7d87bbdd3099f177a3b8b1502b9775a as per https://docs.sentry.io/platforms/go/concurrency/.

Issue #639, opened 2023-08-09
Persistent worker: Configure isolation related parameters in persistent worker config 


anarazel:
When running a persistent worker, particularly for an open source project, it can be important to define the restrictions in the persistent worker config, rather than in a project's .cirrus.yml. Consider e.g. a PR changing `isolation: tart` to `isolation: none`  or changing the tart image to one controlled by the user.

(Originally asked in https://github.com/cirruslabs/cirrus-cli/issues/638#issuecomment-1671949284)

anarazel:
Awesome! Thank you very much.

PR #641, opened 2023-08-15
Introduce Persistent Worker security policy

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/639.


fkorotkov (approved):

internal/commands/worker/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+func TestRestrictOnlySpecificTartImages(t *testing.T) {
+	config, err := parseConfig(filepath.Join("testdata", "security-only-specific-tart-images.yml"))
+	require.NoError(t, err)
+
+	require.Equal(t, []string{"ghcr.io/cirruslabs/*"},
+		config.Security.Isolation.Tart.AllowedImages)
+
+	const goodImage = "ghcr.io/cirruslabs/macos-ventura-base:latest"
+	require.True(t, config.Security.Isolation.Tart.ImageAllowed(goodImage))
+
```
anarazel, line 57: I'd add a test checking that something like `example.org/ghcr.io/cirruslabs/whatnot`` isn't allowed.

internal/executor/instance/persistentworker/persistentworker.go
```diff
@@ -37,11 +56,28 @@ func New(isolation *api.Isolation, logger logger.Lightweight) (abstract.Instance
...
 		return container.New(iso.Container.Image, iso.Container.Cpu, iso.Container.Memory, iso.Container.Volumes)
 	case *api.Isolation_Tart_:
+		tartPolicy := security.TartPolicy()
+		if tartPolicy == nil {
+			return nil, fmt.Errorf("%w: \"tart\" isolation is not allowed by this Persistent Worker's "+
+				"security settings", ErrInvalidIsolation)
+		}
+
+		if !tartPolicy.ImageAllowed(iso.Tart.Image) {
+			return nil, fmt.Errorf("%w: Tart VM image %q is not allowed by this Persistent Worker's "+
```
anarazel, line 74: `Tart` is quoted as `\"tart\"` above. Maybe also add quotes around `%q`?

internal/commands/worker/config_test.go
edigaryev: Good point! Added in https://github.com/cirruslabs/cirrus-cli/pull/641/commits/b977e676c375acc09d41633ed0e4f15df3d30667.

internal/executor/instance/persistentworker/persistentworker.go
edigaryev: You mean change `Tart` to `\"tart\"`? Because `%q` used for the image already [provides quoting](https://pkg.go.dev/fmt).

anarazel: > You mean change Tart to \"tart\"?

Yes.

> Because %q used for the image already [provides quoting](https://pkg.go.dev/fmt).

Sorry - I'm not a go dev at all...

PR #647, opened 2023-09-11
Support YAML flow syntax after expansion

fkorotkov:
This way it's easier to pass lists from Starlark into YAML


edigaryev (approved):

edigaryev:
Rebased against `master` so that failed Windows tests would not cause a build failure.

PR #652, opened 2023-09-21
Executor: cancel the task if no heartbeats were received for the last 2m

edigaryev:
See https://github.com/cirruslabs/cirrus-cli/discussions/643.


internal/executor/executor.go
```diff
@@ -201,9 +202,38 @@ func (e *Executor) runSingleTask(ctx context.Context, task *build.Task) (err err
...
+			case <-ctx.Done():
+				break
+			}
+		}
+	}()
+
+	// Run the task
 	if err := task.Instance.Run(ctx, &instanceRunOpts); err != nil {
 		switch {
+		case errors.Is(ctx.Err(), context.Canceled):
```
fkorotkov, line 230: Would it also be triggered if the parent context is canceled?

edigaryev: Indeed, fixed this in https://github.com/cirruslabs/cirrus-cli/pull/652/commits/bba2f3f03839257e953cadd0a6a8b345f826c0fc.

fkorotkov (approved):

PR #671, opened 2023-12-12
Evaluator: instantiate a filesystem for running Cirrus Hooks

edigaryev:
Let's use the old-way of instantiating a FS from environment variables for now, will switch to https://github.com/cirruslabs/cirrus-cli/pull/629 later.

### DIFF
--- a/internal/evaluator/evaluator.go
+++ b/internal/evaluator/evaluator.go
@@ -200,7 +200,15 @@ func (r *ConfigurationEvaluatorServiceServer) EvaluateFunction(
 	ctx context.Context,
 	request *api.EvaluateFunctionRequest,
 ) (*api.EvaluateFunctionResponse, error) {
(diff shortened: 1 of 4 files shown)


internal/evaluator/evaluator.go
```diff
@@ -282,3 +296,29 @@ func convertFS(apiFS *api.FileSystem) (fs fs.FileSystem, err error) {
...
+	// Use GitHub filesystem if all the required variables are present
+	owner, ok := env["CIRRUS_REPO_OWNER"]
+	if !ok {
+		return
+	}
+	repo, ok := env["CIRRUS_REPO_NAME"]
+	if !ok {
+		return
+	}
+	reference, ok := env["CIRRUS_CHANGE_IN_REPO"]
```
fkorotkov, line 314: This is not reflecting repositories' config resolution policy. It seems we need to pass an FS from the server side.

fkorotkov (approved):

PR #673, opened 2023-12-27
Vetu isolation: support host networking

edigaryev:
See https://github.com/cirruslabs/vetu/pull/35 for more details.


pkg/parser/instance/isolation/vetu/vetu.go
```diff
@@ -90,6 +90,18 @@ func New(mergedEnv map[string]string, parserKit *parserkit.ParserKit) *Vetu {
 		return nil
 	})
 
+	vetu.OptionalField(nameable.NewRegexNameable("networking"), nil, func(node *node.Node) error {
+		if node.HasChild("host") {
```
fkorotkov, line 94: What about bridged?

edigaryev: I don't think we need to expose anything right now as the default networking should be sufficient for `cirrus run` cases.

Note that `networking` here is missing a JSON schema, so it's kinda implies that it's a "hidden" option.

fkorotkov (approved):

PR #677, opened 2024-01-19
Integrate with OpenTelemetry

edigaryev:
With a focus on instrumenting the Persistent Worker for an additional visibility.


fkorotkov (commented):
Is there a way to attache Task ID and Upstream to a context so OTEL will pick it up and annotate accordinally?

fkorotkov (approved):

PR #687, opened 2024-02-14
Store worker tasks in a `SyncMap`

fkorotkov:
Otherwise, we can iterate over it while items get deleted. This caused workers to die with "fatal error: concurrent map iteration and map write".

PS not sure why Sentry didn't get a report so bumped flush timeout to 5 seconds.


fkorotkov:
Not sure about using `github.com/puzpuzpuz/xsync`. Was debating between it and just having an extra mutex.

fkorotkov:
I actually quite like #688. I guess I'll leave it up to you, @edigaryev.

edigaryev (approved):

PR #688, opened 2024-02-14
Concurrency safe worker operations

fkorotkov:
Otherwise, we can iterate over it while items get deleted.

Alternative to #687

### DIFF
--- a/cmd/cirrus/main.go
+++ b/cmd/cirrus/main.go
@@ -36,7 +36,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to initialize Sentry: %v", err)
 	}
-	defer sentry.Flush(2 * time.Second)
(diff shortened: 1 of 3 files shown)


edigaryev:
This PR changes less lines of code indeed, but I think a proper `sync.Map` implementation should incapsulate all of the operations within it, otherwise it's very easy to miss a `Lock()`/`Unlock()`.

This also reminds me of a DIY `ConcurrentMap` implementation in one of the Orchard PRs: https://github.com/cirruslabs/orchard/pull/30/commits/c5079d86ff4297f01b274356b4df7e474aa39bef#diff-10fc4fdc9c214490c8685391c16528854aaaeef9d9c860e6f82709b384d7f657.

I propose to go with the https://github.com/cirruslabs/cirrus-cli/pull/687 to avoid reinventing the wheel.

PR #690, opened 2024-02-26
Persistent Worker: implement standby slots to cache VM startups

edigaryev:
This allows having a cache for popular VM configurations and have a faster startup times.


edigaryev:
Closing because we've decided to simplify things for now:

* only single standby instance is allowed
* only Tart is supported
* no resource mumbo-jumbo, simply kill off the standby instance if its isolation doesn't match the isolation of the task to be started

PR #692, opened 2024-02-28
OpenTelemetry: do not export metrics each second

edigaryev:
When not specified, the interval defaults to 60 seconds.

Should help with the [DPM usage](https://grafana.com/docs/grafana-cloud/cost-management-and-billing/understand-your-invoice/metrics-invoice/#billing-calculations).


fkorotkov (approved):

edigaryev:
For future reference: OpenTelemetry collector still [doesn't seem to support metric aggregation](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/4968).

There's also a mention of DPM in the context of Grafana Cloud in the same issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/4968#issuecomment-1565690792.

Issue #697, opened 2024-03-10
Warm Isolation for Persistent Worker


fkorotkov:
It will be great if before Persistent Worker pull for new tasks it can warm up a preconfigured isolation. I propose to plug this logic in this line:

https://github.com/cirruslabs/cirrus-cli/blob/0b9ed560488909bf0b4802638cb157ac33e825a7/internal/worker/task.go#L56

To have a "warm" isolation configured and see if we need to reuse the warm instance or stop it and create a new one. Worker will also make sure to warm and instance before polling an upstream.

edigaryev:
It seems that this was already implemented in https://github.com/cirruslabs/cirrus-cli/pull/698, so closing.

PR #698, opened 2024-03-11
`WarmableInstance` abstraction

fkorotkov:
Related to #697. I think this will be simplest implementation of the functionality we want.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/tart.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/tart.go
@@ -10,6 +10,7 @@ import (
 	"github.com/cirruslabs/cirrus-cli/internal/executor/instance/runconfig"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/platform"
 	"github.com/cirruslabs/cirrus-cli/internal/logger"
+	"github.com/cirruslabs/echelon"
 	"github.com/getsentry/sentry-go"
 	"github.com/google/uuid"
 	"github.com/samber/lo"
@@ -46,7 +47,7 @@ type Tart struct {
 	display     string
 	volumes     []*api.Isolation_Tart_Volume
 
-	mountTemporaryWorkingDirectoryFromHost bool
+	vm *VM
 }
 
 func New(
@@ -78,7 +79,11 @@ func New(
 	return tart, nil
 }
 
-func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err error) {
+func (tart *Tart) Warmup(ctx context.Context, additionalEnvironment map[string]string, logger *echelon.Logger) error {
+	return tart.bootVM(ctx, additionalEnvironment, "", logger)
+}
+
+func (tart *Tart) bootVM(ctx context.Context, additionalEnvironment map[string]string, automountDir string, logger *echelon.Logger) error {
 	ctx, prepareInstanceSpan := tracer.Start(ctx, "prepare-instance")
 	defer prepareInstanceSpan.End()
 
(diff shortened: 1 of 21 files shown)


internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -88,53 +93,26 @@ func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err err
 		})
 	}
 
-	tmpVMName := fmt.Sprintf("%s%d-", vmNamePrefix, config.TaskID) + uuid.NewString()
+	tmpVMName := fmt.Sprintf("%s-warm-", vmNamePrefix) + uuid.NewString()
```
edigaryev, line 96: This seems to add the `-warm-` string to *all* Tart instances and not just the warm ones.

Should we add an additional argument to `Warmup()` that will specify this optional string to be added to VM names in https://github.com/cirruslabs/cirrus-cli/pull/700?

fkorotkov: Good idea. Maybe allow to optinally specify `vmNamePrefix` as `"cirrus-warm"` vs the default `"cirrus"`?

edigaryev: Added in https://github.com/cirruslabs/cirrus-cli/pull/700/commits/b5590f927f7a17ba0c6fc2586a82aa04a45e80eb.

internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -192,22 +188,47 @@ func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err err
...
 
 	bootLogger.Errorf("VM was assigned with %s IP", ip)
 	bootLogger.Finish(true)
 
+	return nil
+}
+
+func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err error) {
+	if tart.vm != nil && config.DirtyMode {
+		return fmt.Errorf("%w: dirty mode is not supported for a warmed instance", ErrFailed)
```
edigaryev, line 203: Not all Tart instances will be standby instances (this assumption was also in https://github.com/cirruslabs/cirrus-cli/pull/698#discussion_r1523350920).

I think we still have a `cirrus run --dirty` (https://github.com/cirruslabs/cirrus-cli/issues/554) or are we dropping it for Tart?

internal/executor/instance/persistentworker/isolation/tart/tart.go
```diff
@@ -192,22 +188,47 @@ func (tart *Tart) Run(ctx context.Context, config *runconfig.RunConfig) (err err
 		break
 	}
 
+	tart.vm = vm
+
 	tart.logger.Debugf("IP %s retrieved from VM %s, running agent...", ip, vm.Ident())
```
edigaryev, line 193: I think this should be in `Run()`, we're not running the agent here.

internal/executor/instance/persistentworker/isolation/tart/tart.go
fkorotkov: There is no dirty mode for persistent workers so this should never happen. Warmable instances won't be used for `cirrus run`

internal/executor/instance/persistentworker/isolation/tart/tart.go
fkorotkov: Moved in [04fee80](https://github.com/cirruslabs/cirrus-cli/pull/698/commits/04fee80eef05acaff472362efe61466cd1985116)

edigaryev (approved):

PR #700, opened 2024-03-13
Persistent Worker: support running a single standby instance

edigaryev:
Now using the `WarmableInstance` abstraction from https://github.com/cirruslabs/cirrus-cli/pull/698.

### DIFF
--- a/internal/worker/task.go
+++ b/internal/worker/task.go
@@ -101,7 +101,7 @@ func (worker *Worker) getInstance(ctx context.Context, isolation *api.Isolation)
 		worker.logger.Debugf("standby instance does not match the task's isolation configuration, " +
 			"terminating it")
 
-		if err := worker.standbyInstance.Close(ctx); err != nil {
+		if err := standbyInstance.Close(ctx); err != nil {
 			worker.logger.Errorf("failed to terminate the standby instance: %v", err)
 		} else {
 			worker.logger.Debugf("standby instance had successfully terminated")
--- a/internal/worker/task.go
+++ b/internal/worker/task.go
@@ -15,6 +15,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/proto"
 	"maps"
 	"net/url"
 	"runtime"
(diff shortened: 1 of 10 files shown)


internal/worker/worker.go
```diff
@@ -196,6 +206,51 @@ func (worker *Worker) Run(ctx context.Context) error {
...
+		return
+	}
+
+	// Do nothing if the standby instance is already instantiated
+	if worker.standbyInstance != nil {
+		return
+	}
+
+	// Do nothing if there are tasks that are running to simplify the resource management
+	if worker.tasks.Size() != 0 {
```
fkorotkov, line 221: I think it's OK for now but we'll need to think how to make a standby if we run one VM already.

internal/worker/worker.go
```diff
@@ -196,6 +206,51 @@ func (worker *Worker) Run(ctx context.Context) error {
...
+		worker.logger.Errorf("failed to create a standby instance: %v", err)
+
+		return
+	}
+
+	logger := echelon.NewLogger(echelon.TraceLevel, renderers.NewSimpleRenderer(os.Stdout, nil))
+
+	worker.logger.Debugf("warming-up the standby instance")
+
+	if err := standbyInstance.(abstract.WarmableInstance).Warmup(ctx, nil, logger); err != nil {
```
fkorotkov, line 238: Why not to use `worker.logger`?

internal/worker/task.go
```diff
@@ -82,6 +83,35 @@ func (worker *Worker) startTask(
 	worker.logger.Infof("started task %d", agentAwareTask.TaskId)
 }
 
+func (worker *Worker) getInstance(ctx context.Context, isolation *api.Isolation) (abstract.Instance, error) {
```
fkorotkov, line 86: Let's add some OTEL metric on hit rate.

internal/worker/task.go
```diff
@@ -82,6 +83,35 @@ func (worker *Worker) startTask(
...
+				"yielding it to the task")
+
+			return standbyInstance, nil
+		}
+
+		// Otherwise terminate the standby instance to simplify the resource management
+		worker.logger.Debugf("standby instance does not match the task's isolation configuration, " +
+			"terminating it")
+
+		if err := worker.standbyInstance.Close(ctx); err != nil {
```
fkorotkov, line 104: `worker.standbyInstance` is `nil` here already

edigaryev: Good catch, fixed in https://github.com/cirruslabs/cirrus-cli/pull/700/commits/364b6f2a3906d8fa21733760606a6a522a95e14a.

internal/worker/worker.go
edigaryev: Because the `Warmup()` expects an `*echelon.Logger`, which is not what we use for the worker.

internal/worker/task.go
edigaryev: Added in https://github.com/cirruslabs/cirrus-cli/pull/700/commits/62c104a0b90737ad0ce238f0b3e3238cc77c2c26.

PR #701, opened 2024-03-21
Standby improvements

fkorotkov:
Respect resources and added tests

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -50,8 +50,8 @@ type Worker struct {
 
 	logger logrus.FieldLogger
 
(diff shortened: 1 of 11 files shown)


internal/worker/worker.go
```diff
@@ -171,7 +171,7 @@ func (worker *Worker) Run(ctx context.Context) error {
 		return err
 	}
 
-	// Standby-related metrics
+	// StandbyConfig-related metrics
```
edigaryev, line 174:
```suggestion
	// Standby-related metrics
```

edigaryev (approved):

PR #702, opened 2024-03-23
Fixed standby example

fkorotkov:


PERSISTENT-WORKERS.md
```diff
@@ -303,7 +303,7 @@ persistent_worker:
       user: admin
       password: admin
       cpu: 4
-      memory: 12
+      memory: 12288 # 13 GB
```
edigaryev, line 306: Why do we need this?

`12` here gets correctly parsed as `12288` by [`ParseMegaBytes()`](https://github.com/cirruslabs/cirrus-cli/blob/ba524650bd2eaefbc3b7ba59b641ffff689d8774/pkg/parser/instance/resources/resources.go#L18).

Issue #733, opened 2024-05-20
Allow test selection for `cirrus internal test`


fkorotkov:
Right now all tests will be executed. There should be a way to filter out tests and/or choose root of the folder for self-discovery.

PR #735, opened 2024-05-20
cirrus internal test: allow filtering the tests using a glob expression

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/733.

### DIFF
--- a/cmd/cirrus/main.go
+++ b/cmd/cirrus/main.go
@@ -3,10 +3,12 @@ package main
 import (
 	"context"
 	"crypto/x509"
+	"errors"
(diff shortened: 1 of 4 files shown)


fkorotkov (commented):
Should we fail if nothing matched?

edigaryev:
> Should we fail if nothing matched?

Good point, implemented in https://github.com/cirruslabs/cirrus-cli/pull/735/commits/504b3082e212d1018dd3a0ea8ddf55bc623f3bd5.

fkorotkov (approved):

PR #741, opened 2024-06-18
Sign macOS Binaries

fkorotkov:
So far it's failing so it's a WIP


fkorotkov:
Seems something weird was with the initial certificate. Regenerating seems to help.

PR #760, opened 2024-07-23
RPC & persistent worker: support string-based task IDs

edigaryev:
Once the backend will start sending and receiving task IDs as strings, we can fully remove the `int64` support from CLI.


fkorotkov:
Why can't we stick with ints? String was created for Cirrus Teams but we tabled it for now.

PR #764, opened 2024-08-06
Persistent Worker: introduce resource modifiers

edigaryev:
Things that are still left to be implemented:

* integration test

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/vetu/vetu.go
+++ b/internal/executor/instance/persistentworker/isolation/vetu/vetu.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cirruslabs/cirrus-cli/internal/executor/instance/runconfig"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/platform"
 	"github.com/cirruslabs/cirrus-cli/internal/logger"
+	"github.com/cirruslabs/cirrus-cli/internal/worker/resourcemodifier"
 	"github.com/cirruslabs/echelon"
 	"github.com/getsentry/sentry-go"
 	"github.com/google/uuid"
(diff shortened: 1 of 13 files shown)


internal/executor/instance/persistentworker/isolation/vetu/vm.go
```diff
@@ -131,6 +136,11 @@ func (vm *VM) Start(
 			args = append(args, "--net-host")
 		}
 
+		// Apply "run" resource modifier
+		if vm.resourceModifier != nil {
```
fkorotkov, line 140: Maybe just extra caution?

```suggestion
		if vm.resourceModifier != nil && vm.resourceModifier.Append != nil {
```

internal/executor/instance/persistentworker/isolation/vetu/vetu.go
```diff
@@ -190,6 +194,10 @@ func (vetu *Vetu) WorkingDirectory(projectDir string, dirtyMode bool) string {
 }
 
 func (vetu *Vetu) Close(ctx context.Context) error {
+	if vetu.resourceModifier != nil {
+		vetu.resourceModifier.Unlock()
```
fkorotkov, line 198: should it be deferred? So once the VM is closed aka stopped we release the resource? Otherwise there might be a race.

internal/executor/instance/persistentworker/isolation/vetu/vm.go
edigaryev: This won't compile because the `Append` can't be `nil` 🤔

internal/executor/instance/persistentworker/isolation/vetu/vetu.go
edigaryev: Good catch, added `defer` in https://github.com/cirruslabs/cirrus-cli/pull/764/commits/ce9524a187415698dec64064ec9f5b7e2f214300.

fkorotkov (approved):

Issue #766, opened 2024-08-11
[worker] `podman system service` process accumulation


duckinator:
When running `cirrus worker run` with the container isolation on Debian 12, with the Podman backend, each task spawns a `podman system service` process that never exits.

At one point, the system had accumulated 158 of these processes, but hadn't run any tasks for over 3 hours.

**WORKAROUND:** Periodically close and re-opening `cirrus worker run`.

duckinator:
This line is what runs the process that doesn't exit:

https://github.com/cirruslabs/cirrus-cli/blob/95e9f68f96873e4d9f8d9bad2f6ce6208aef491d/internal/executor/instance/containerbackend/podman_linux.go#L43

However, my understanding is that this function should eventually be called to kill it:

https://github.com/cirruslabs/cirrus-cli/blob/95e9f68f96873e4d9f8d9bad2f6ce6208aef491d/internal/executor/instance/containerbackend/podman_linux.go#L92-L121

I'm unsure whether this function isn't being called, or if it's just not working for some reason.

duckinator:
I changed `"-t", "0"` to `"-t", "1"` and it spawns zombie processes instead.

This confirms that, for whatever reason, the `podman system service` processes are not being waited on.

edigaryev:
I've tried reproducing your issue on Cirrus CLI 0.122.0 and Podman 3.4.4 on a clean `ghcr.io/cirruslabs/ubuntu:latest` instance with the following configuration to no avail:

```yaml
container:
  image: debian:latest

task:
  script: uname -a
```

https://github.com/cirruslabs/cirrus-cli/pull/767 might help, though.

edigaryev:
I'm running it via [Tart](https://github.com/cirruslabs/tart) on macOS, but that probably doesn't matter much for this reproduce attempt.

edigaryev:
@duckinator how does your `.cirrus.yml` look like? Which Cirrus CLI and Podman versions do you run and on which Linux distribution?

We'll probably need to reproduce this somehow first in order to devise a fix (if it's needed at all).

Looking at the code it should cleanup an instance or throw an error:

https://github.com/cirruslabs/cirrus-cli/blob/2a07c0284c95b1eec626b7e8341fa96ed3585ebe/internal/executor/executor.go#L208-L217

duckinator:
Versions and such:
* distro: Debian 12 (bookworm)
* cirrus: 0.122.0-95e9f68
[...]

PR #767, opened 2024-08-12
Set WaitDelay for the spawned "podman system service" process

edigaryev:
Resolves https://github.com/cirruslabs/cirrus-cli/issues/766.


fkorotkov:
Why does it help with the reported issue?

edigaryev:
Hmm, I've just realized that we don't set the `{Stdin,Stdout,Stderr}` fields of the `*exec.Cmd` and https://github.com/golang/go/issues/23019 won't help in this case.

Closing.

PR #774, opened 2024-08-27
cirrus_ci_service.proto: support multipart cache uploads

edigaryev:
To go beyond the 5 GB limit and upload parts in parallel.


api/cirrus_ci_service.proto
```diff
@@ -264,6 +264,49 @@ service CirrusCIService {
...
+  string upload_id = 1;
+}
+
+message MultipartCacheUploadPartRequest {
+  CacheKey cache_key = 1;
+  string upload_id = 2;
+  uint32 part_number = 3;
+}
+
+message MultipartCacheUploadPartResponse {
```
fkorotkov, line 288: Can it be `GenerateURLResponse`?

edigaryev: `GenerateURLResponse` does not account for multiple values in a single header, which pre-signing S3 client does.

Also, this this seems to go against the [Create Unique Protos per Method](https://protobuf.dev/programming-guides/api/#unique-protos) best practice.

fkorotkov: I'm just asking because `GenerateURLResponse` contains a pre-signed URL and we don't encounter multiple values per header situation.

edigaryev: See https://github.com/cirruslabs/cirrus-cli/pull/774/commits/a013f13f3e6103db2565abddf1d3b5082bd4bc4d.

fkorotkov (approved):

edigaryev:
Added `content_length` field to `MultipartCacheUploadPartRequest` as it's required by R2, see https://github.com/cirruslabs/chacha/pull/4.

fkorotkov (approved):

PR #777, opened 2024-08-30
Pre-pull Vetu VM before running tests

edigaryev:
Could help with TestWorkerStandByVetu timing out after 10 minutes.


edigaryev:
Actually, the culprit was in the https://github.com/cirruslabs/vetu/pull/58, but I think this change is still nice to have.

fkorotkov (approved):

PR #789, opened 2024-09-19
GHA cache: retry temporary failures on part uploads and commits

fkorotkov:
By propagating proper status and using StatusServiceUnavailable which is re-triable by the GHA client.

Plus made the parts commit re-triable by not marking parts as finalized.

https://github.com/actions/toolkit/blob/6dd369c0e648ed58d0ead326cf2426906ea86401/packages/cache/src/internal/requestUtils.ts#L24-L34

### DIFF
--- a/internal/agent/http_cache/ghacache/ghacache.go
+++ b/internal/agent/http_cache/ghacache/ghacache.go
@@ -79,7 +79,7 @@ func (cache *GHACache) get(writer http.ResponseWriter, request *http.Request) {
 			return
 		}
 
-		fail(writer, request, http.StatusInternalServerError, "GHA cache failed to "+
+		fail(writer, request, http.StatusServiceUnavailable, "GHA cache failed to "+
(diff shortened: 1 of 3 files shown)


edigaryev:
I'm not sure if *any* error should be retried.

The default behavior is [2 retries with 5000 ms timeout](https://github.com/actions/toolkit/blob/6dd369c0e648ed58d0ead326cf2426906ea86401/packages/cache/src/internal/requestUtils.ts#L99-L100), so that's a *minimum* of 10 seconds of retrying things that possibly make no sense to retry.

Another issue is idempotency.

How about we start with a less invasive option in https://github.com/cirruslabs/cirrus-cli/pull/788, add Sentry error logging on top of it, and go from there?

fkorotkov:
Closing in favor of https://github.com/cirruslabs/cirrus-cli/pull/788 after applying some middle ground fixes.

PR #790, opened 2024-09-23
GHA cache: report failures to Sentry

edigaryev:
And also move towards structured logging a bit more by parametrizing the failure attributes.

Looks like this (note the `url`, which [`github.com/getsentry/sentry-go/http`](https://pkg.go.dev/github.com/getsentry/sentry-go/http) package provides:

Screenshot 2024-09-23 at 21 01 31

And this:

Screenshot 2024-09-23 at 21 01 15

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/docker/docker v27.2.0+incompatible
 	github.com/docker/go-units v0.5.0
 	github.com/dustin/go-humanize v1.0.1
-	github.com/getsentry/sentry-go v0.27.0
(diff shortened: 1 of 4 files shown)


fkorotkov (approved):
I think we need to make sure we pass SENTRY_DNS to the agent and make sure the agent is not leaking it to script instructions.

edigaryev:
Let's merge this as is for now, I'll follow-up with either a (1) `SENTRY_DSN` passthrough or (2) a way to collect these logs in Persistent Worker and tag them with a task ID and other tags.

PR #791, opened 2024-09-25
Migrate to string IDs

fkorotkov:
This change migrates tasks ids to strings so we can support UUIDs for a secret project. This change tries to make it in a backward compatible way:

* Agent part will continue to provide the `TaskIdentification` proto in case `-task-id` is a parsable integer.
* Persistent Worker part continues to support both old integer ids and the new strings ids.
  * in case string id is empty it falls back to the old
  * reports running tasks ids in strings but parsable ints in the old proto field


internal/executor/executor.go
```diff
@@ -85,7 +85,7 @@ func New(projectDir string, tasks []*api.Task, opts ...Option) (*Executor, error
 
 			// Lowest priority: task-specific
 			environment.NodeInfo(task.LocalGroupId, int64(len(tasks))),
-			environment.TaskInfo(task.Name, task.LocalGroupId),
+			environment.TaskInfo(task.Name),
```
fkorotkov, line 88: It was weird that it used `LocalGroupId`. Seems like a bug that was overwritten anyways after merging with other envs.

edigaryev (approved):

PR #814, opened 2024-11-26
Introduce macOS "Local Network" permission helper process

edigaryev:
And implement privilege separation and dropping, which can be achieved by passing the `--user` command-line argument to `cirrus worker run`.

Closes https://github.com/cirruslabs/cirrus-cli/pull/816.

### DIFF
--- a/internal/commands/localnetworkhelper.go
+++ b/internal/commands/localnetworkhelper.go
@@ -1,0 +1,26 @@
+package commands
+
+import (
+	"github.com/cirruslabs/cirrus-cli/pkg/localnetworkhelper"
+	"github.com/spf13/cobra"
+)
+
+func newLocalNetworkHelperCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    localnetworkhelper.CommandName,
+		Short:  "Run the macOS \"Local Network\" permission helper process",
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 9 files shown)


edigaryev:
Some pre-text and reasoning:

1. The ["Local Network" permission](https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy) introduced in macOS 15 (Sequoia) effectively requires running Cirrus CLI as `root` in order to be able to access Tart VMs over SSH without the GUI permission pop-up (which can't be disabled or automated)

3. However, it feels that it's worth pursuing a higher security target than just running the whole Cirrus CLI process as `root`, if possible

4. One way to achieve that is to implement a classic [privelege separation](https://en.wikipedia.org/wiki/Privilege_separation), creating a Berkeley socket in the privileged (helper) process and sending it to the unprivileged (main) process using [FD passing technique](https://pkg.go.dev/github.com/markdingo/go-fdpass-demo)

5. Another way, implemented in this PR, is to establish a persistent SSH session between the two processes, and simply use [`*ssh.Client`](https://pkg.go.dev/golang.org/x/crypto/ssh#Client) methods like [`Dial()`](https://pkg.go.dev/golang.org/x/crypto/ssh#Client.Dial) and [`DialContext()`](https://pkg.go.dev/golang.org/x/crypto/ssh#Client.DialContext), which utilize SSH protocol TCP/IP forwarding and replicate their [`net` counterpart](https://pkg.go.dev/net) behavior

PR #824, opened 2025-01-13
Dynamically configurable standby instance

fkorotkov:
Moved standby configuration in proto and added configuration to gRPC polling methods.

### DIFF
--- a/internal/executor/instance/persistentworker/isolation/tart/tart.go
+++ b/internal/executor/instance/persistentworker/isolation/tart/tart.go
@@ -137,7 +137,8 @@ func (tart *Tart) Warmup(
 	var monitorCtx context.Context
 	var monitorCancel context.CancelFunc
 	if warmup.TimeoutSeconds != 0 {
-		monitorCtx, monitorCancel = context.WithTimeoutCause(ctx, time.Duration(warmup.TimeoutSeconds)*time.Second, abstract.ErrWarmupTimeout)
+		duration := time.Duration(warmup.TimeoutSeconds) * time.Second
+		monitorCtx, monitorCancel = context.WithTimeoutCause(ctx, duration, abstract.ErrWarmupTimeout)
 	} else {
 		monitorCtx, monitorCancel = context.WithCancel(ctx)
 	}


api/cirrus_ci_service.proto
```diff
@@ -158,6 +158,7 @@ message PollRequest {
   repeated int64 old_running_tasks = 2 [deprecated = true];
   map<string, double> resources_in_use = 3;
   repeated string running_tasks = 4;
+  StandbyInstanceInformation availableStandbyInstanceInformation = 5;
```
edigaryev, line 161: We should probably use snake case here too, to be consistent with the rest of the file.

internal/worker/worker.go
```diff
@@ -303,8 +304,9 @@ func (worker *Worker) tryCreateStandby(ctx context.Context) {
 
 	worker.logger.Debugf("warming-up the standby instance")
 
+	warmUpTimeout := time.Duration(worker.standbyParameters.Warmup.TimeoutSeconds) * time.Second
```
edigaryev, line 307: This seems to cause NPE in tests: https://cirrus-ci.com/task/5965693838950400?logs=test#L3291-L3303

internal/worker/standby.go
```diff
@@ -39,22 +30,21 @@ func (standby *StandbyConfig) UnmarshalYAML(value *yaml.Node) error {
...
 		return ErrIsolationMissing
 	}
 	// Parse isolation
 	parserKit := &parserkit.ParserKit{
 		Boolevator:    boolevator.New(),
 		IssueRegistry: issue.NewRegistry(),
 	}
-	isolationParser := isolation.NewIsolation(nil, parserKit)
-	if err := isolationParser.Parse(isolationNode, parserKit); err != nil {
+	instanceParser := instance.NewStandbyParameters(nil, parserKit)
```
edigaryev, line 41: Should this be renamed to avoid confusion? We don't parse an instance here, but standby parameters.

I think just naming it `parser` should do the trick.

internal/worker/worker.go
edigaryev: Another point is that since the `worker.standbyParameters` and its inner messages are now `nil`-able, and the standby parameters can be set from the backend, it would be nice if we can check for `nil` here in a more defensive manner.

edigaryev (approved):

PR #828, opened 2025-01-21
[tuist] don't validate security requirements

fkorotkov:
Since we are running a local server proxy which handles security differently.

### DIFF
--- a/internal/agent/http_cache/tuistcache/tuistcache_test.go
+++ b/internal/agent/http_cache/tuistcache/tuistcache_test.go
@@ -16,16 +16,6 @@ import (
 	"testing"
 )
 
-type NoSecurity struct{}
-
-func (NoSecurity) Authorization(_ context.Context, _ string) (tuistapi.Authorization, error) {
-	return tuistapi.Authorization{}, nil
-}
-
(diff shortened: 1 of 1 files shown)


edigaryev:
But we already use "no security": https://github.com/cirruslabs/cirrus-cli/blob/8006bef8c7d1961d6dfc1b607bd3b5f4a051e47e/internal/agent/http_cache/tuistcache/tuistcache.go#L38

We have a test that doesn't use any security either and it passes. What does this PR this change in this regard? 🤔

edigaryev:
> Will take a look tomorrow why existing test passes.

Thank you.

It feels to me that there's an easier way to get around this, for example, by returning an `ogenerrors.ErrSkipClientSecurity` from our `NoSecurity` implementation.

edigaryev:
OK, so this error:

```
{"error_message":"operation DownloadCacheArtifact: security \"\": security requirement is not satisfied"}
```

Is coming from this template, from `ogenerrors.SecurityError{}`: https://github.com/ogen-go/ogen/blob/c120ddb415a27ba87ba33b8b4a849edf19b93414/gen/_template/handlers.tmpl#L131-L195.

It seems that there's no way to disable this via `ogen.yaml`.

edigaryev (approved):

PR #841, opened 2025-02-17
Agent(targz): make symlink targets relative to their current directory

edigaryev:
This way symbolic links pointing to a path that starts with a base directory will continue to work after `Unarchive()`, because these symbolic links will point to location relative to their own location.

Right now they are made relative to the base directory, which is only correct when such a link resides in that base directory.


edigaryev:
> This is not a regression though?

I don't think so.

PR #846, opened 2025-03-07
Back off if failed to create a standby

fkorotkov:
Otherwise, we'll just proceed to polling without a warm standby.


edigaryev:
> Otherwise, we'll just proceed to polling without a warm standby.

We'll try to create a standby once more on the next poll cycle, no?

PR #847, opened 2025-03-10
Log blob proxying progress

fkorotkov:
Every minute in case proxying is taking longer than one minute.

Hope it will help with understanding the issues with R2 when downloads timeout.

### DIFF
--- a/internal/agent/http_cache/azureblob/getblob.go
+++ b/internal/agent/http_cache/azureblob/getblob.go
@@ -3,7 +3,9 @@ package azureblob
 import (
 	"fmt"
 	"github.com/cirruslabs/cirrus-cli/internal/agent/client"
+	"github.com/cirruslabs/cirrus-cli/internal/agent/progressreader"
 	"github.com/cirruslabs/cirrus-cli/pkg/api"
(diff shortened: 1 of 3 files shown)


internal/agent/http_cache/azureblob/getblob.go
```diff
@@ -76,14 +78,33 @@ func (azureBlob *AzureBlob) getBlob(writer http.ResponseWriter, request *http.Re
...
 	// we usually proxy large files so let's use a larger buffer
 	largeBuffer := make([]byte, PROXY_DOWNLOAD_BUFFER_SIZE)
-	bytesRead, err := io.CopyBuffer(writer, resp.Body, largeBuffer)
-	if err != nil {
-		proxyingDuration := time.Since(startProxyingAt)
-		fail(writer, request, http.StatusInternalServerError, "failed to proxy cache entry download",
-			"err", err, "duration", proxyingDuration, "read", bytesRead, "key", key)
-
-		return
+	bytesRead := int64(0)
```
edigaryev, line 84: What do you think about moving all of this into a separate function that we'll call instead of `io.CopyBuffer`?

PR #897, opened 2025-05-01
Support `optimistically_restore_on_miss`

fkorotkov:
This will prefix match a cache entry and download the latest available for a given instruction.

### DIFF
--- a/internal/executor/rpc/cache.go
+++ b/internal/executor/rpc/cache.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"github.com/cirruslabs/cirrus-cli/internal/executor/cache"
 	"github.com/cirruslabs/cirrus-cli/pkg/api"
+	"github.com/samber/lo"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"io"
@@ -128,13 +129,19 @@ func (r *RPC) CacheInfo(ctx context.Context, req *api.CacheInfoRequest) (*api.Ca
 
 	var file *os.File
 
+	var prefixMatch bool
+
 	if req.CacheKey != "" {
 		file, err = r.build.Cache.Get(req.CacheKey)
 	}
-	for _, prefix := range req.CacheKeyPrefixes {
-		file, err = r.build.Cache.FindByPrefix(prefix)
-		if file != nil {
-			break
+	if file == nil {
+		for _, prefix := range req.CacheKeyPrefixes {
+			file, err = r.build.Cache.FindByPrefix(prefix)
+			if file != nil {
+				prefixMatch = true
+
+				break
+			}
 		}
 	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/agent/executor/cache.go
```diff
@@ -114,6 +115,15 @@ func (executor *Executor) DownloadCache(
 
 	cachePopulated, cacheAvailable := executor.tryToDownloadAndPopulateCache(ctx, logUploader, commandName, cacheHost, cacheKey, baseFolder)
 
+	if !cacheAvailable && instruction.OptimisticallyRestoreOnMiss {
+		logUploader.Write([]byte("\nWasn't able to find the exact cache! Requesting the last available one..."))
+		cacheInfo := executor.findLatestAvailableCache(ctx, logUploader, commandName)
+		if cacheInfo != nil {
```
edigaryev, line 121: It would probably be a good idea to handle `cacheInfo == nil` to and print a log entry explaining what happened.

internal/agent/executor/cache.go
```diff
@@ -194,6 +204,27 @@ func (executor *Executor) generateCacheKey(
 	return fmt.Sprintf("%s-%x", commandName, cacheKeyHash.Sum(nil)), true
 }
 
+func (executor *Executor) findLatestAvailableCache(ctx context.Context, uploader *LogUploader, commandName string) *api.CacheInfo {
+	cacheInfoRequest := api.CacheInfoRequest{
+		TaskIdentification: client.CirrusTaskIdentification,
+		CacheKey:           commandName + "-", // Use a prefix that will match any cache key for this command
```
edigaryev, line 210: Should this use `CacheKeyPrefixes` instead of `CacheKey` if we want prefix matching to be used?

internal/agent/executor/cache.go
```diff
@@ -194,6 +204,27 @@ func (executor *Executor) generateCacheKey(
...
+
+	if err != nil {
+		uploader.Write([]byte(fmt.Sprintf("\nFailed to find latest available cache for %s: %v\n", commandName, err)))
+		return nil
+	}
+	if response != nil && response.Info != nil {
+		return response.Info
+	}
+
+	uploader.Write([]byte(fmt.Sprintf("\nThere is no other cache entry available for %s: %v\n", commandName)))
```
edigaryev, line 223:
```suggestion
	uploader.Write([]byte(fmt.Sprintf("\nThere is no other cache entry available for %s\n", commandName)))
```

internal/executor/rpc/cache.go
```diff
@@ -125,7 +126,17 @@ func (r *RPC) CacheInfo(ctx context.Context, req *api.CacheInfoRequest) (*api.Ca
...
 	r.logger.Debugf("sending info about cache key %s", req.CacheKey)
 
-	file, err := r.build.Cache.Get(req.CacheKey)
+	var file *os.File
+
+	if req.CacheKey != "" {
+		file, err = r.build.Cache.Get(req.CacheKey)
+	}
+	for _, prefix := range req.CacheKeyPrefixes {
+		file, err = r.build.Cache.FindByPrefix(prefix)
```
edigaryev, line 135: There's a potential `*os.File` leak here, but given the ephemeral nature of `cirrus run`, we could probably just ignore it.

edigaryev: Should be fixed by https://github.com/cirruslabs/cirrus-cli/pull/899.

edigaryev:
It seems that `CacheInfo()` currently returns an empty key when prefix matching was utilized, should be fixed by https://github.com/cirruslabs/cirrus-cli/pull/899.

edigaryev (approved):

PR #907, opened 2025-06-03
Reset `dockerfile*` fields in case `image` is set

fkorotkov:
To not have an issue where both are specified.

### DIFF
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -71,7 +71,19 @@ func TestValidConfigs(t *testing.T) {
 	for _, validCase := range validCases {
 		file := validCase
 		t.Run(file, func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/parser/instance/container.go
```diff
@@ -34,16 +35,6 @@ func NewCommunityContainer(
 		},
 	}
 
-	imageSchema := schema.String("Docker Image to use.")
```
edigaryev, line 37: Is there a reason to move this and other `imageSchema`'s below in their corresponding files?

edigaryev: I've reverted the order in https://github.com/cirruslabs/cirrus-cli/pull/908 and the tests still pass.

The PR above explains why this happens.

PR #921, opened 2025-06-27
ghacachev2: do not escape cache key when not necessary

edigaryev:
We already have `url.PathEscape()` in `azureBlobURL()`.

Otherwise [this part of `github.com/tonistiigi/go-actions-cache`](https://github.com/moby/buildkit/blob/a23bc16feff9789f207a7b59220ce79c86444a39/vendor/github.com/tonistiigi/go-actions-cache/cache.go#L440-L443) fails.


edigaryev:
Also added support for matching by key prefixes in `cirruscimock`'s `CacheInfo()` implementation in https://github.com/cirruslabs/cirrus-cli/pull/921/commits/43f5e193b39780f816d97b1e5eb09ff65595b2d2.

This way it's possible to point the `docker/build-push-action` to a local `http_cache` instance (using ngrok, etc.) and see the cache working.

PR #943, opened 2025-08-25
Agent: do not limit extendedTimeoutCtx to 5 minutes

edigaryev:
To fix regression introduced in https://github.com/cirruslabs/cirrus-cli/pull/938.


edigaryev:
Nope, this doesn't seem to be correct either: if we cancel the task, `on_timeout:` or `always:` block might run much more than 5 minutes.

PR #958, opened 2025-10-16
azureblob: recover from io.ErrUnexpectedEOF whenever possible

edigaryev:
Sometimes we get `io.ErrUnexpectedEOF` from the upstream, but the downstream might not implement re-tries.

Use `Range` and `If-Range` headers to read the rest of the data.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -84,7 +84,7 @@ task:
     - export PATH=$PATH:/usr/local/go/bin
     - go test -v -p 1 ./...
   env:
-    CIRRUS_INTERNAL_VETU_VM: ghcr.io/cirruslabs/ubuntu-runner-arm64:latest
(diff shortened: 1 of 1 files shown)


internal/agent/http_cache/azureblob/getblob.go
```diff
@@ -132,19 +145,104 @@ func (azureBlob *AzureBlob) proxyCacheEntryDownload(
...
+			return 0, fmt.Errorf("range start + bytes read (%d) is larger than range end (%d)",
+				start+bytesRead, *end)
+		}
+
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", start+bytesRead, *end))
+	} else {
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-", start+bytesRead))
+	}
+
+	req.Header.Set("If-Range", ifRangeValue)
```
fkorotkov: Nice. TIL.

fkorotkov (approved):

PR #965, opened 2025-11-07
Abstract HTTP Cache

fkorotkov:


edigaryev:
> Will try to split work in many PRs given that it's unclear how to support caching with pre-signed URLs for local `cirrus run`.

Maybe we can just keep the `UploadCache()` and `DownloadCache()` in `cirrus_ci_service.proto` for now?

This will solve the problem and allow us to move forward with this PR.

PR #1017, opened 2026-02-09
Add eval command

fkorotkov:


internal/commands/eval.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/cirruslabs/cirrus-cli/pkg/larker/loader"
+	"github.com/spf13/cobra"
+	"go.starlark.net/starlark"
+)
+
+var ErrEval = errors.New("eval failed")
+
+func newEvalCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "eval <script.star|->",
```
edigaryev: What do you think about making this command accept standard input by default, and make the file optional?

```suggestion
		Use:   "eval [FILE]",
```

Similarly to how `python` works, but without too much interactivity for now.

internal/commands/eval.go
```diff
@@ -0,0 +1,134 @@
...
+
+This command is a lightweight, LLM-friendly way to run Python-like scripts.
+Instead of starting a full Python interpreter, it executes Starlark directly
+and exposes Cirrus built-ins that are useful for automation and data handling.
+
+Load built-ins like this:
+  load("cirrus", "http", "fs", "json", "yaml")
+
+Output is produced only by print(...) and println(...).`,
+		Example: strings.TrimSpace(`
```
edigaryev: This example feels a bit too noisy in the command's help (`cirrus eval --help`).

Are you sure we need it?

fkorotkov: Implemented in c613846. I simplified the help examples to concise file/stdin usage and removed the long heredoc + HTTP example to reduce noise in `cirrus eval --help`.

